PR #2, opened 2021-04-28
support for atlas query, status update, OLM basics

jeremyary:


jeremyary:
@jianrongzhang89 going to go ahead and merge up so we can move forward with work we need to get through for the week, but an FYI this is out here if we want to have some convo on it.

wtrocki:
@jeremyary is this runablle right now? Can I take look

jeremyary:
@wtrocki It is! I've got it deployed and running on our lab, happy to show you there on a quick vid chat or google chat me and I can walk you thru make commands and what resource you'll need to create to trigger it if you'd like. The functionality is not fully there yet for creating the resource that we'll be "binding" against, though that's coming soon. Thus far, it's all about querying Atlas and getting instance info back to allow the admin to select from in a mock UI.

wtrocki:
Thank you. Will check code and we can have some Dev sync later

PR #6, opened 2021-05-01
DBAAS-21/24/25: allow DBaaSConn to bind to apps

jeremyary:
## Description
[DBAAS-21: DBaaSConnection CRD](https://issues.redhat.com/browse/DBAAS-21)
[DBAAS-24: DBaaSConnection creates owned deployment](https://issues.redhat.com/browse/DBAAS-24)
[DBAAS-25: binding annotations for DBaaSConnection](https://issues.redhat.com/browse/DBAAS-25)

I've done some refactoring to pull resource details into a `models` package in addition to having our DBaaSConnection create a bindable deployment with a sample configmap/secret for connection info that can be injected into a busybox app for verification.

## Verification Steps
Will fill this in Monday so it can be verified by others

## Type of change

- [X] New feature (non-breaking change which adds functionality)

- [X] All acceptance criteria specified in JIRA or in issue number have been completed
- [ ] Code Review completed


bundle/manifests/dbaas.redhat.com_dbaasconnections.yaml
```diff
@@ -3,6 +3,11 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.4.1
+    service.binding/db.host: path={.status.dbConfigMap},objectType=ConfigMap
+    service.binding/db.name: path={.status.dbConfigMap},objectType=ConfigMap
+    service.binding/db.password: path={.status.dbCredentials},objectType=Secret
+    service.binding/db.port: path={.status.dbConfigMap},objectType=ConfigMap
```
wtrocki, line 9: This doesn't look like correct reference

config/samples/dbaas_v1_dbaasservice.yaml
```diff
@@ -8,25 +8,25 @@ spec:
...
-        - cloudProvider: TENANT
-          cloudRegion: US_EAST_1
-          id: 608484560795100e26a40caa
-          instanceSizeName: M0
-          name: Cluster0
-      id: 6084842759a7bd6d9b0ea0e8
-      name: example project
-      users:
-        - name: dbExampleUser
+# Once Atlas has been queried based on the credential secret auth info provided, results will populate status -
```
wtrocki, line 11: My recommendation at early phases is to allow unknown fields and do not define status.
This will save you troubles with some updates without bumping crd version

controllers/dbaasconnection_controller.go
```diff
@@ -37,27 +42,117 @@ type DBaaSConnectionReconciler struct {
...
+		}
+	}
 
+	// DBaaSConnection resource found, try and find matching bindable deployment
+	existingDeployment := models.Deployment(&dbaasConnection)
+	err = r.Get(ctx, client.ObjectKeyFromObject(existingDeployment), existingDeployment)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// matching Deployment has not yet been created, make one with ownerReference
+			newDeployment := models.OwnedDeployment(&dbaasConnection)
```
wtrocki, line 75: So it is hacked then :)

config/crd/patches/service_binding_annotations.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: dbaasconnections.dbaas.redhat.com
+  annotations:
+    service.binding/db.name: 'path={.status.dbConfigMap},objectType=ConfigMap'
```
wtrocki, line 6: Missing type and provider from spec. If target is to use quarkus keys have been different if I recall

config/samples/dbaas_v1_dbaasservice.yaml
bmozaffa: A clarification that applies especially to this comment, but also potentially to other comments, is that we are currently working on a 1-2 week demo sprint where we will be hardcoding many things and only focus on functional outcome.

wtrocki: Yes. I'm aware. Not defining status would be advice that fits hardcoding everything, model. Not sure how I read this. Is it like 
1 `no need to review things`? or 2 `those comments are irrelevant at this phase`?

bmozaffa: Feedback is always welcome, and it's neither.

We've had lengthy debates about how strictly defined CRDs should be in other projects, and there are certainly pros and cons. I expect a similar debate for this project, and we should/will have it. I'm just saying there is plenty of time. We will continue to update the CRD while making incompatible changes and without respecting the version increments, for a while.

config/crd/patches/service_binding_annotations.yaml
bmozaffa: Wojciech, I wasn't able to find pre-defined keys in Quarkus, would you please point us to them?

Semi-relatedly, I'm not sure if it was you who said service binding as files was the preferred approach and "supported" by Quarkus. Do you have a reference or any details on how Quarkus reads the mounted files to form a database connection?

wtrocki: Yes.. problem is that they are not defined anywhere (I'm trying to address that problem), but for the moment you can look on binding converters for different dbs.

https://github.com/quarkusio/quarkus/search?q=ServiceBindingConverter

I would love to get your comment on this issue

https://github.com/quarkusio/quarkus/issues/16737

bmozaffa: Thanks, but it looks like it's only JDBC (and Kafka, etc) so far, whereas our first target is MongoDB! Can we plug into the ServiceBindingConverter framework, and would you happen to have an example of that (which is not a quarkus module)?

wtrocki: Service binding values will be there anyway. So you can:

1) hardcode values from binding in application.properties
2) write your own transformer (that would be trivial considering that examples for jdbc are there
3) use jdbc for mongo https://docs.mongodb.com/datalake/tutorial/jdbc-driver/
4) create issue in Quarkus to support this (recommending this anyway)
5) detect binding but not use values from it (works for the demo ;) )

wtrocki: Another option that works best with mongo is to use node.js

This library/poc will give you binding much faster etc..

https://github.com/nodeshift/kube-service-bindings/issues

We can discuss with @mhdawson if you would like to get out of the box mongo support for it later

bmozaffa: Thanks, Wojciech, this works: [https://github.com/bmozaffa/mongodb-quarkus](https://github.com/bmozaffa/mongodb-quarkus/blob/main/src/main/java/org/acme/mongodb/servicebinding/MongoServiceBindingConverter.java)

jeremyary: type/provider added

bundle/manifests/dbaas.redhat.com_dbaasconnections.yaml
jeremyary: Do you mean just the type/provider that was missing from the other comment or something else here?

jeremyary:
Merging up so we can get started on actual field values coming back from Atlas. ATM, this breaks the binding over spec/provider due to the new annotation, but I'll have code up today/tomorrow to fix this and get legitimate values into our binding annotation targets rather than mock data.

bundle/manifests/dbaas.redhat.com_dbaasconnections.yaml
wtrocki: Only those type to be consistent with spec

config/crd/patches/service_binding_annotations.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: dbaasconnections.dbaas.redhat.com
+  annotations:
+    service.binding/type: 'path={.spec.type}'
```
wtrocki, line 6: In future those would have been specified here as static fields.
We work on this. Provider is just for info reasons and it usually is your operator name

jeremyary: thanks, I'll update provider in next round of changes

PR #16, opened 2021-07-05
DBAAS-79: use a config map and a secret for all binding parameters

xieshenzh:
## Description

Use a config map and a secret for all binding parameters, automatically mapped with their key/values when mounted on application.
Jira: https://issues.redhat.com/browse/DBAAS-79

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

- [ ] Code Review completed


xieshenzh:
@bmozaffa This PR is for the API design changes related to the service binding parameters.

api/v1alpha1/dbaasprovider.go
```diff
@@ -110,12 +107,12 @@ type DBaaSConnectionSpec struct {
...
-	// The connection string for this instance
-	ConnectionString string `json:"connectionString,omitempty"`
-
-	// Secret holding username and password
+	// Secret holding credentials needed for accessing the DB instance
 	CredentialsRef *corev1.LocalObjectReference `json:"credentialsRef,omitempty"`
 
+	// ConfigMap holding non-sensitive information needed for connecting to the DB instance
+	ConfigMapRef *corev1.LocalObjectReference `json:"configMapRef,omitempty"`
+
```
bmozaffa, line 115: Avishay combined these two parameters into a single ConnectionInfo in the API ADR, please update again based on that.

config/crd/patches/service_binding_annotations.yaml
```diff
@@ -3,9 +3,6 @@ kind: CustomResourceDefinition
...
   name: dbaasconnections.dbaas.redhat.com
   annotations:
-    service.binding/type: 'path={.spec.type}'
-    service.binding/provider: 'path={.spec.provider}'
-    service.binding/database: 'path={.status.dbConfigMap},objectType=ConfigMap'
-    service.binding/host: 'path={.status.dbConfigMap},objectType=ConfigMap'
-    service.binding/username: 'path={.status.dbCredentials},objectType=Secret'
-    service.binding/password: 'path={.status.dbCredentials},objectType=Secret'
\ No newline at end of file
+    service.binding/provider: dbaas-operator
```
bmozaffa, line 6: Should this be the database operator? Not sure if it has a functional impact, but does it allow the application to determine if it's, for example, Crunchy postgres or Cockroach postgres?

bmozaffa (commented):
Thanks, Xieshen!

config/crd/patches/service_binding_annotations.yaml
xieshenzh: According to the API ADR, the value of `service.binding/provider` is fixed to “dbaas-operator”.  
@jeremyary do you know what the value should be?

jeremyary: The response received is this:
```
Even the spec says that "provider is an entry with a value that
identifies the provider of the binding". Now, I could imagine that
this info can be used by frameworks to tune accessing the service, but
the entry if fully optional.

TBH, we did not establish any particular conventions, and your
experience would be very valuable if we want to go in that direction.
Looking forward to your feedback.
```

we originally used 'dbass-operator' as suggested by Woj early on for demo and this seems as acceptable as any else since there's nothing looking at this field other than maybe manually troubleshooting a cluster. We create the binding object, I think our value makes sense.

jeremyary: Since this is now pushed down into provider ConfigMap, shouldn't hold up merging AFAICT

PR #15, opened 2021-07-05
DBAAS-69: watch DBaaSProvider inventory and connection CRs 

xieshenzh:
## Description

Watch the provider ConfigMaps changes and re-deploy DBaaS operator
Jira: https://issues.redhat.com/browse/DBAAS-69

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA or in issue number have been completed
- [ ] Unit tests added that prove the fix is effective or the feature works 
- [ ] Documentation added for the feature
- [ ] all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer

PR #25, opened 2021-07-23
DBAAS-128: create CRD for DBaaSProvider

xieshenzh:
## Description

Create CRD for DBaaSProvider.
Jira: https://issues.redhat.com/browse/DBAAS-128

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

- [ ] Code Review completed


xieshenzh:
I'll update the DBaaSProvider controller to reload the operator for provider registration in PR https://github.com/RHEcosystemAppEng/dbaas-operator/pull/15 .

api/v1alpha1/dbaasprovider_types.go
```diff
@@ -0,0 +1,52 @@
...
+//+kubebuilder:resource:scope=Cluster
+
+//+operator-sdk:csv:customresourcedefinitions:displayName="DBaaSProvider"
+// DBaaSProvider is the Schema for the dbaasproviders API
+type DBaaSProvider struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   DBaaSProviderSpec   `json:"spec,omitempty"`
+	Status DBaaSProviderStatus `json:"status,omitempty"`
```
xieshenzh, line 38: It seems we cannot create an API without `status`, because operator-sdk [reads the status when generating CRD](https://github.com/operator-framework/operator-sdk/blob/a450af0f981e86e52abb0d2e5d4b46b729e55bc2/internal/generate/clusterserviceversion/bases/definitions/crd.go#L109).
So I kept an empty status for DBaaSProvider.

redhatHameed (approved):
LGTM

api/v1alpha1/dbaasprovider.go
```diff
@@ -21,46 +21,68 @@ import (
...
+
+	// Icon information indicates what logo we display on developer catalog tile
+	Icon ProviderIcon `json:"icon"`
 }
 
-type AuthenticationField struct {
-	Name   string `json:"name"`
-	Masked bool   `json:"masked"`
+// The Kind of the provider API
+type Kind string
```
jeremyary, line 54: Thinking here... Avishay was recommending we try with the legitimate API 'Kind' type, but when I look up TypeMeta, it also gives Kind as a String. Unless you think it makes sense for this operator to use something like a schema.GroupKind (and I'm not even sure that's the one we'd even want), maybe we leave those two fields as strings instead of declaring this new struct that just holds the string instead. 

WDYT? The way you use it for unstructured stuff, what object type for these makes the most sense for you?

api/v1alpha1/dbaasprovider.go
```diff
@@ -21,46 +21,68 @@ import (
...
+
+// ProviderIcon follows same field/naming formats as CSV
+type ProviderIcon struct {
+	Data      string `json:"data"`
```
jeremyary, line 58: json field was designed as `base64data` to align with [CSV values](https://github.com/xieshenzh/dbaas-operator/blob/main/config/manifests/bases/dbaas-operator.clusterserviceversion.yaml#L31)

api/v1alpha1/dbaasprovider.go
```diff
@@ -80,7 +104,7 @@ type Instance struct {
 }
 
 type NamespacedName struct {
-	// The namespace where object of known type is store
+	// The namespace where object of known type is stored
 	Namespace string `json:"namespace,omitempty"`
```
jeremyary, line 108: Though I see now that it's laid out explicitly in the API doc (which I believe is so that it's clear to the reader what's represented), we shouldn't need to redefine this type in code. Use from "k8s.io/apimachinery/pkg/types"

config/samples/dbaas_v1alpha1_dbaasprovider.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: dbaas.redhat.com/v1alpha1
+kind: DBaaSProvider
+metadata:
+  name: dbaasprovider-sample
+spec:
+  # Add fields here
+  foo: bar
```
jeremyary, line 7: I don't think this is what the legit object looks like :) Should mostly be able to use my example from [here](https://gist.github.com/jeremyary/48ab828bc03bffaeada143c051c63505) if you'd like with maybe a few tweaks.

api/v1alpha1/dbaasprovider.go
redhatHameed: +1 was thinking same, 	as it get convert it into string-> 		`Kind:    string(provider.Spec.InventoryKind)` , `Kind: string(provider.Spec.ConnectionKind),`

api/v1alpha1/dbaasprovider.go
xieshenzh: The `NamespacedName` from "k8s.io/apimachinery/pkg/types" doesn't have json tags.
Can we use [ObjectReference](https://github.com/kubernetes/api/blob/aa30bdaf750c49387bf90f9c4f01d29bf6ee56e0/core/v1/types.go#L5405) from `k8s.io/api/core/v1` instead?

jeremyary: ahhh right, ok... sorry, I guess you had the right approach then in defining it. ObjectReference seems pretty heavy-handed for our use case since we just need the few fields.

bmozaffa: I created our own namespace name for that reason, so you can use that, if it's still there

bmozaffa: Sorry, replied to this comment without reading the whole thread

api/v1alpha1/dbaasprovider.go
xieshenzh: I thinks `string` is the better option. Thanks.

api/v1alpha1/dbaasprovider.go
xieshenzh: I changed the json tag to `data` because in case we want to unmarshall the data with `mapstructure`, there will be problem if the json tag doesn't match the field name.
Now that I replaced `mapstructure` with `encoding/json`, there won't be any problem to use `base64data`.
Thanks.

jeremyary:
Sounds good. Let me know when you're ready and I'll review again and get a merge going. Thanks for your efforts on this!

xieshenzh:
@jeremyary I have validated the changes. It is ready for review. Thanks.

controllers/suite_test.go
```diff
@@ -67,6 +67,9 @@ var _ = BeforeSuite(func() {
 	err = dbaasv1alpha1.AddToScheme(scheme.Scheme)
 	Expect(err).NotTo(HaveOccurred())
 
+	err = dbaasv1alpha1.AddToScheme(scheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
```
jeremyary, line 72: this seems to be duplication of the lines directly above it?

jeremyary (commented):
one small question/change noted in comments. Will get us merged up right after that. Thanks!!

controllers/suite_test.go
xieshenzh: Corrected, thanks.

PR #26, opened 2021-07-28
open DBaaSProvider view to all auth'd users

jeremyary:
## Description
Add view permissions for DBaaSProvider resources for all authenticated users so that the UX can fetch & list providers regardless of the current namespace. Refer to security ADR for further info under 'Tenant Registration'.

## Type of change

- [X] New feature (non-breaking change which adds functionality)


xieshenzh (approved):
Looks good to me. Thanks.

redhatHameed (approved):

PR #19, opened 2021-07-13
DBAAS-104: add webhook validation for instance ID & InventoryRef on DBaaSConnectionSpec to immutable

xieshenzh:
## Description

Add webhook validation for instance ID & InventoryRef on DBaaSConnectionSpec to immutable
Jira: https://issues.redhat.com/browse/DBAAS-104

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA or in issue number have been completed
- [ ] Unit tests added that prove the fix is effective or the feature works 
- [ ] Documentation added for the feature
- [ ] all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer

PR #28, opened 2021-08-08
[DBAAS-136] add DBaaSTenant CRD and support multi-tenancy

tchughesiv:

## Description
Add ability to manage DBaaS user security, for multiple tenants and personas, in any Openshift cluster.
 - new DBaaSTenant object and controller
 - default Tenant created on operator startup to support single-tenancy
 - access controls provided for a Tenant's Service Admins and Developers

## Type of change

- [x] Documentation added for the feature
- [ ] Code Review completed


config/crd/patches/webhook_in_dbaastenants.yaml
```diff
@@ -0,0 +1,16 @@
...
+  conversion:
+    strategy: Webhook
+    webhook:
+      clientConfig:
+        service:
+          namespace: system
+          name: webhook-service
+          path: /convert
+      conversionReviewVersions:
+      - v1
```
jeremyary, line 16: I haven't done a ton with webhooks yet, so I was [reading](https://olm.operatorframework.io/docs/advanced-tasks/adding-admission-and-conversion-webhooks/#conversion-webhook-rules-requirements) on this type. It mentions a required field on the CRD:
```
The CRD targeted by the Conversion webhook must have its spec.preserveUnknownFields field set to false or nil
```
I don't see it on our CRD, but I do see L41 in Makefile that seems to pass this as a CRD option. Does this take care of the same thing in a broader approach encompassing all CRDs instead?

jeremyary (dismissed):
Tests pass for me, code looks good. A non-blocker question & comment to your question inline, otherwise lgtm. Thanks!

config/crd/patches/webhook_in_dbaastenants.yaml
jeremyary: Ok, makes sense. Thanks for explaining. We have introduced a couple webhooks for making a couple fields, in essence, immutable: 
https://github.com/RHEcosystemAppEng/dbaas-operator/pull/19

jeremyary (approved):

config/crd/patches/webhook_in_dbaastenants.yaml
tchughesiv: nice

xieshenzh: The webhook in the file is a conversion webhook.
The webhook from #19 is a validating webhook.

tchughesiv: @xieshenzh true... however, we're not currently using this webhook for tenants. this is a standard file created by the sdk for all new crds. it will be available to us if we require it in the future.

controllers/dbaastenant_controller.go
```diff
@@ -0,0 +1,306 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var (
+	TenantList        v1alpha1.DBaaSTenantList
+	TenantNames       []string
+	TenantInventoryNS []string
```
xieshenzh, line 45: The variables are potentially shared by multiple goruntines. 
Ideally, if an Inventory needs to know its Tenant, we should find a way to link the Inventory to its Tenant, and let the inventory reconciler read the tenant directly.
Otherwise, we need to consider synchronization when sharing the variables.

controllers/dbaastenant_controller.go
```diff
@@ -0,0 +1,306 @@
...
+			}
+			logger.Info(clusterRoleBindingObj.Kind+" resource updated", "Name", clusterRoleBindingObj.Name)
+		}
+	}
+
+	// Reconcile each inventory in the tenant's namespace to ensure proper RBAC is created
+	for _, inventory := range inventoryList.Items {
+		// should we return anything on err for these reconciles?
+		// _, err = r.InventoryCtrl.Reconcile(ctx, invReq)
+		r.InventoryCtrl.Reconcile(ctx, reconcile.Request{
```
xieshenzh, line 136: I think we could find a way to trigger Reconcile and let it handle errors, instead of invoking the method directly.

controllers/dbaasinventory_controller.go
```diff
@@ -188,56 +187,30 @@ func inventoryRbacObjs(inventory v1alpha1.DBaaSInventory) (rbacv1.Role, rbacv1.R
...
-		}
-		for _, group := range inventory.Spec.Authz.Groups {
-			roleBinding.Subjects = append(roleBinding.Subjects, getSubject(group, role.Namespace, "Group"))
+	roleBinding.SetGroupVersionKind(rbacv1.SchemeGroupVersion.WithKind("RoleBinding"))
+
+	// if inventory.Spec.Authz is nil, use tenant defaults for view access to the Inventory object
+	var users, groups []string
+	if inventory.Spec.Authz.Users == nil && inventory.Spec.Authz.Groups == nil {
+		for _, tenant := range TenantList.Items {
+			if tenant.Spec.InventoryNamespace == inventory.Namespace {
```
xieshenzh, line 196: Do we allow different tenants to use the same namespace, or we need to find a way to prevent this?

xieshenzh (approved):
It looks good for the default tenant. Added some questions for future multi-tenancy work. Thanks!

controllers/dbaastenant_controller.go
tchughesiv: certainly worth exploring... we just need to be sure that every inventory, in a given tenant's namespace, is reconciled whenever a tenant is created/modified

controllers/dbaasinventory_controller.go
tchughesiv: currently, we do allow this and there's no reason to prevent it. as part of this PR, i've added some unit tests which cover that very use-case.

controllers/dbaastenant_controller.go
tchughesiv: > The variables are potentially shared by multiple goruntines.
yes, that was the intention... to have a current list of tenants made available to other the controllers

> Ideally, if an Inventory needs to know its Tenant, we should find a way to link the Inventory to its Tenant, and let the inventory reconciler read the tenant directly.


> Otherwise, we need to consider synchronization when sharing the variables.
definitely open to suggestions. as it's currently designed... any time a tenant is created or modified, these vars are updated. then, as part of the same tenant controller... every inventory in that tenant's namespace will have it's reconcile method triggered.

PR #29, opened 2021-08-10
update to show dbaasinventory instead of dbaasservice

cdfurgal:
## Description
updating dbaasservice to say dbaasinventory


jeremyary (approved):
lgtm!

PR #48, opened 2021-09-23
going to 0.1.1 with candidate-0.1 channel

jeremyary:
## Description
Bumping to v0.1.1 along with a change to align with delivery best practices and start delivering via a `candidate-0.1` channel for incremental changes to our alpha release.


config/samples/quarkus-runner.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: v1
```
redhatHameed, line 1: can we ada a crunchy example?

redhatHameed: also can we change version here -> https://github.com/RHEcosystemAppEng/dbaas-operator/blob/main/config/samples/catalog-source.yaml#L8

jeremyary: Didn't really mean to include but they're not a bad idea to have available :) Added crunchy sample & renamed both to be more obvious what we're supplying.

Version in CatalogSource updated as well.

redhatHameed (dismissed):

bmozaffa:
This is a little ironic, because the fruit app was originally developed for jdbc and I ported it to Mongo... so why not use the same thing for psql?

jeremyary:
Changing the channel back to alpha as apparently Managed Tenants doesn't support the recommended channel structure:
```
13:29:28 [AddonLoadError] /var/lib/jenkins/workspace/service-managed-tenants-gl-pr-check/addons/dbaas-operator validation error: 'candidate-0.1' is not one of ['alpha', 'beta', 'stable', 'edge', 'rc']
```

bmozaffa:
> This is a little ironic, because the fruit app was originally developed for jdbc and I ported it to Mongo... so why not use the same thing for psql?

Actually nvm, mixing up with something.  But check, is there not the same sample for jdbc?

config/samples/quarkus-crunchydata-sample-app.yaml
```diff
@@ -0,0 +1,50 @@
...
+          name: postgresql-orm-quarkus
+      template:
+        metadata:
+          labels:
+            name: postgresql-orm-quarkus
+        spec:
+          containers:
+            - name: postgresql-orm-quarkus
+              # Replace this with the built image name
+              image: quay.io/ecosystem-appeng/postgresql-orm-quarkus:v0.1
```
bmozaffa, line 21: @jeremyary Is this the same as hibernate-orm-quickstart?

redhatHameed: yes @bmozaffa

redhatHameed: copied from https://github.com/quarkusio/quarkus-quickstarts/tree/main/hibernate-orm-quickstart

xieshenzh (approved):

PR #53, opened 2021-10-05
switch to ubi image

tchughesiv:


tchughesiv:
RH only offers a 1.15 go-toolset image, but we've been using 1.16. there appears to be a conflict when builting w/ 1.15 which i'm still working through.

tchughesiv:
until the official rh go-toolset image offers go 1.16, we'll install the necessary version ourselves during the build process.

.github/workflows/dockerbuild.yml
```diff
@@ -0,0 +1,21 @@
+name: Build Image
```
jeremyary, line 1: is this action just meant to attempt the build and notify if an issue occurs?

tchughesiv: yes

PR #63, opened 2021-10-18
[DBAAS-161] implicitly determine service admins

tchughesiv:
## Description
Implicitly determine the service admin(s) for a Tenant. With this PR, if a user, group, or serviceaccount has permission to "create" both `dbaasinventories` and `secrets` in the tenant's namespace, they are considered a service admin for that tenant.

Similarly, with this change, we implicitly determine the developer persona for some users... specifically those with "list" `dbaasinventories` permissions in a tenant namespace. 

This is done with the use of `resourceaccessreviews` and removes the need for a `tenant.spec.authz.serviceAdmin` field.

When watching `rolebindings` for changes, only the metadata is cached to cut down on memory consumption.

An additional printer column has been added to indicate the inventory namespace for tenants -
```shell
$ oc get dbaastenant cluster
NAME      INVENTORY_NS               AGE
cluster   openshift-dbaas-operator   17m
```

## Type of change

- [ ] Code Review completed


tchughesiv:
@jianrongzhang89 reading your comment again, i think i understand your concern now. we won't be giving anyone those create inventory/secret permissions. instead, we do a lookup to see who already has those rights and determine that those users/groups are service admins for that tenant. all this really means for our operator is that those users should be given get access to that tenant object

that's all we do, our operator doesn't grant any user or group anything beyond `read` perms on a cluster object.

jianrongzhang89 (approved):

controllers/dbaasinventory_controller.go
```diff
@@ -139,9 +135,22 @@ func (r *DBaaSInventoryReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	return ctrl.NewControllerManagedBy(mgr).
+	c, err := ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.DBaaSInventory{}).
 		Build(r)
+	if err != nil {
+		return nil, err
+	}
+
+	// watch rolebindings
+	if err = c.Watch(
```
xieshenzh, line 146: I think it is better to create a new controller for `RoleBinding`, and keep this inventory controller only reconciling for one `Inventory` at a time. Otherwise, when an inventory is updated, all the inventories will be reconciled.

controllers/dbaastenant_controller.go
```diff
@@ -226,34 +232,29 @@ func tenantRbacObjs(tenant v1alpha1.DBaaSTenant, inventoryAuthz v1alpha1.DBaasUs
...
-		clusterRoleBinding.Subjects = append(clusterRoleBinding.Subjects, getSubject(group, "", "Group"))
-	}
+	clusterRoleBinding.Subjects = getSubjects(users, groups, "")
 
 	return clusterRole, clusterRoleBinding
 }
 
 // get cumulative authz from all inventories in namespace
-func getAllAuthzFromInventoryList(inventoryList v1alpha1.DBaaSInventoryList, tenant v1alpha1.DBaaSTenant) (inventoryAuthz v1alpha1.DBaasUsersGroups) {
+func getDevAuthzFromInventoryList(inventoryList v1alpha1.DBaaSInventoryList, tenant v1alpha1.DBaaSTenant) (developerAuthz v1alpha1.DBaasUsersGroups) {
```
xieshenzh, line 244: This function could be moved to `dbaas_reconciler.go` as it is used by multiple reconcilers.

controllers/dbaasinventory_controller.go
tchughesiv: agreed

tchughesiv: done

tchughesiv: good call, and doing so allowed for simplification of the inventory and tenant reconcilers

tchughesiv: i've decided instead to move everything authz/rbac related to the tenant controller. This is because the tenant object is the main driver of authz and the entire reason the tenant object exists is for this purpose. The tenant controller will now reconcile by watching tenant and rolebinding objects in the cluster.

jeremyary (dismissed):

controllers/dbaastenant_controller.go
```diff
@@ -121,49 +103,52 @@ func (r *DBaaSTenantReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		GenericFunc: func(e event.GenericEvent) bool {
-			return false
+			return checkForDBaasOwnership(e.Object)
 		},
 	}
 }
 
 // create a default Tenant if one doesn't exist
-func (r *DBaaSTenantReconciler) createDefaultTenant(ctx context.Context, tenantList v1alpha1.DBaaSTenantList) (ctrl.Result, error) {
+func (r *DBaaSTenantReconciler) createDefaultTenant(ctx context.Context) (ctrl.Result, error) {
```
jianrongzhang89, line 137: Should the default tenant have an owner reference set? Since the default tenant is created by the DBaaS operator, it would be nice to have it cleaned up when the operator itself is removed.

xieshenzh (approved):

controllers/dbaastenant_authorization.go
```diff
@@ -0,0 +1,444 @@
...
+
+	developerAuthz := r.getDeveloperAuthz(ctx, tenant, inventoryList)
+	serviceAdminAuthz := r.getServiceAdminAuthz(ctx, tenant)
+	clusterRole, clusterRolebinding := tenantRbacObjs(tenant, serviceAdminAuthz, developerAuthz)
+	var clusterRoleObj rbacv1.ClusterRole
+	if exists, err := r.createRbacObj(&clusterRole, &clusterRoleObj, &tenant, ctx); err != nil {
+		return err
+	} else if exists {
+		if !reflect.DeepEqual(clusterRole.Rules, clusterRoleObj.Rules) {
+			clusterRoleObj.Rules = clusterRole.Rules
```
jianrongzhang89, line 159: Could this lead to dangling rule objects in the cluster if clusterRoleObj.Rules contains more entries than clusterRole.Rules?

controllers/dbaastenant_authorization.go
```diff
@@ -0,0 +1,444 @@
...
+			logger.Info(clusterRoleObj.Kind+" resource updated", "Name", clusterRoleObj.Name)
+		}
+	}
+	var clusterRoleBindingObj rbacv1.ClusterRoleBinding
+	if exists, err := r.createRbacObj(&clusterRolebinding, &clusterRoleBindingObj, &tenant, ctx); err != nil {
+		return err
+	} else if exists {
+		if !reflect.DeepEqual(clusterRolebinding.RoleRef, clusterRoleBindingObj.RoleRef) ||
+			!reflect.DeepEqual(clusterRolebinding.Subjects, clusterRoleBindingObj.Subjects) {
+			clusterRoleBindingObj.RoleRef = clusterRolebinding.RoleRef
```
jianrongzhang89, line 173: Same question as above.

controllers/dbaastenant_controller.go
```diff
@@ -20,86 +20,67 @@ import (
...
-	// get list of DBaaSTenants
-	tenantList := v1alpha1.DBaaSTenantList{}
-	if err := r.List(ctx, &tenantList); err != nil {
-		logger.Error(err, "Error fetching DBaaS Tenant List for reconcile")
-		return ctrl.Result{}, err
-	}
-
 	// create default tenant if none exists
-	if result, err := r.createDefaultTenant(ctx, tenantList); err != nil {
+	if result, err := r.createDefaultTenant(ctx); err != nil {
```
jianrongzhang89, line 58: Should the reconciler update the tenant CR status so that the user knows whether the CR has been reconciled successfully?

controllers/dbaastenant_controller.go
tchughesiv: for the same reason we don't remove inventories, we shouldn't remove tenant objects. the operator being uninstalled doesn't mean the existing tenants/inventories/connections aren't still in use.

also, an accidental/temporary operator removal shouldn't remove a user's environment. Similarly, OLM doesn't delete an operator's CRDs when an operator is uninstalled

jianrongzhang89: Sounds good.

controllers/dbaastenant_controller.go
```diff
@@ -196,64 +182,21 @@ func getDefaultTenant(inventoryNamespace string) v1alpha1.DBaaSTenant {
...
-
-	for _, user := range users {
-		clusterRoleBinding.Subjects = append(clusterRoleBinding.Subjects, getSubject(user, "", "User"))
-	}
-	for _, group := range groups {
-		clusterRoleBinding.Subjects = append(clusterRoleBinding.Subjects, getSubject(group, "", "Group"))
-	}
-
-	return clusterRole, clusterRoleBinding
+	return tenantNames
```
jianrongzhang89, line 190: Since a named return parameter is used, it would be better to return without value.

controllers/dbaastenant_authorization.go
tchughesiv: we're replacing the rules in `clusterRoleObj` entirely with the ones in `clusterRole`

controllers/dbaastenant_controller.go
tchughesiv: done

controllers/dbaastenant_controller.go
tchughesiv: possibly yes, but i think we should address this in a separate PR. could you put some initial thoughts together on what that might look like?

jeremyary (approved):

PR #64, opened 2021-10-19
DBAAS-203: reduce the permissions that the operator needs for platform reconciler

xieshenzh:
## Description

Disable the uninstallation of the platform resources and reduce the permissions that the operator needs.

Jira: https://issues.redhat.com/browse/DBAAS-203

## Verification Steps

1. Install the dbaas operator
2. Check if the platform resources are installed in the same namespace as the dbaas operator
3. Uninstall the dbaas operator
4. Check if the namespace-scoped platform resources are all deleted
5. Re-install the dbaas operator 
6. Check if everything is up and running

## Type of change

- [ ] Code Review completed

### DIFF
--- a/controllers/dbaasplatform_controller.go
+++ b/controllers/dbaasplatform_controller.go
@@ -24,29 +24,49 @@ import (
 	"strings"
 	"time"
 
+	dbaasv1alpha1 "github.com/RHEcosystemAppEng/dbaas-operator/api/v1alpha1"
 	"github.com/RHEcosystemAppEng/dbaas-operator/controllers/reconcilers"
 	"github.com/RHEcosystemAppEng/dbaas-operator/controllers/reconcilers/console_plugin"
 	"github.com/RHEcosystemAppEng/dbaas-operator/controllers/reconcilers/crunchybridge_installation"
 	"github.com/RHEcosystemAppEng/dbaas-operator/controllers/reconcilers/csv"
 	"github.com/RHEcosystemAppEng/dbaas-operator/controllers/reconcilers/mongodb_atlas_instalation"
 	"github.com/RHEcosystemAppEng/dbaas-operator/controllers/reconcilers/servicebinding"
+
 	"github.com/go-logr/logr"
+
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
-
-	dbaasv1alpha1 "github.com/RHEcosystemAppEng/dbaas-operator/api/v1alpha1"
 )
 
+/*
+Platform resources to be created or updated explicitly by the DBaaS operator:
+	Crunchy Bridge operator: CatalogSource, Subscription, OperatorGroup
+	MongoDB Atlas operator: CatalogSource, Subscription, OperatorGroup
+	Service Binding operator: Subscription
+	DBaaS Dynamic Plugin: Service, Deployment, ConsolePlugin, Console (updated)
+	Console Telemetry Plugin: Service, Deployment, ConsolePlugin, Console (updated)
+
+Platform resources to be removed by the DBaaS operator by setting owner reference:
+	Crunchy Bridge operator: Subscription, CSV
+	MongoDB Atlas operator: Subscription, CSV
+	Service Binding operator: Subscription, CSV
+	DBaaS Dynamic Plugin: Service, Deployment
+	Console Telemetry Plugin: Service, Deployment
+
+Platform resources NOT to be removed or updated by the DBaaS operator:
+	Crunchy Bridge operator: CatalogSource (in different namespace), OperatorGroup (in different namespace)
+	MongoDB Atlas operator: CatalogSource (in different namespace), OperatorGroup (in different namespace)
+	DBaaS Dynamic Plugin: ConsolePlugin (cluster-scoped), Console (cluster-scoped)
+	Console Telemetry Plugin: ConsolePlugin (cluster-scoped), Console (cluster-scoped)
+*/
+
 const (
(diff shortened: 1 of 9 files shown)


bmozaffa:
Thanks for this! Does this end up cleaning up via owner references, or does it, as I suggested, just not fully clean up to avoid the perms? If we face a choice between cleaning up and not having elevated permissions, we should probably go with the latter, but let's double check before making that decision. Relatedly, since OSD uses addons, does the addon uninstall not have the ability to clean up outside of our operator code?

controllers/dbaasplatform_controller.go
```diff
@@ -273,12 +272,12 @@ func (r *DBaaSPlatformReconciler) getReconcilerForPlatform(provider dbaasv1alpha
 		return mongodb_atlas_instalation.NewReconciler(r.Client, r.Scheme, r.Log)
 	case dbaasv1alpha1.DBassDynamicPluginInstallation:
 		return console_plugin.NewReconciler(r.Client, r.Log,
-			reconcilers.DBAAS_DYNAMIC_PLUGIN_NAME, reconcilers.DBAAS_DYNAMIC_PLUGIN_NAMESPACE,
+			reconcilers.DBAAS_DYNAMIC_PLUGIN_NAME, r.InstallNamespace,
```
jianrongzhang89, line 275: So here the plugins will be installed in the dbaas operator's namespace. Should the SBO and provider operators get installed in this namespace too?

tchughesiv: I think, definitely the provider operators should be considered for this... but with SBO, why aren't we installing that as a csv dependency? aren't we using the official SBO operator?

tchughesiv (approved):
lgtm

controllers/dbaasplatform_controller.go
```diff
@@ -60,42 +58,35 @@ type DBaaSPlatformReconciler struct {
 //+kubebuilder:rbac:groups=dbaas.redhat.com,resources=dbaasplatforms,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=dbaas.redhat.com,resources=dbaasplatforms/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=dbaas.redhat.com,resources=dbaasplatforms/finalizers,verbs=update
-//+kubebuilder:rbac:groups=operators.coreos.com,resources=catalogsources;subscriptions;operatorgroups;clusterserviceversions,verbs=get;list;create;update;delete;watch
-//+kubebuilder:rbac:groups=apps,resources=deployments;daemonsets;statefulsets,verbs=get;list;create;update;delete;watch
-//+kubebuilder:rbac:groups=core,resources=namespaces;services,verbs=get;list;create;update;delete;watch
-//+kubebuilder:rbac:groups=console.openshift.io,resources=consoleplugins,verbs=get;list;create;update;delete;watch
+//+kubebuilder:rbac:groups=operators.coreos.com,resources=catalogsources;operatorgroups,verbs=get;list;create;update;watch
+//+kubebuilder:rbac:groups=operators.coreos.com,resources=subscriptions;clusterserviceversions,verbs=get;list;create;update;watch;delete
```
tchughesiv, line 62: do we still require `delete` perms for csv/subs? this seems unnecessary?

tchughesiv (changes requested):

controllers/dbaasplatform_controller.go
xieshenzh: The owner references of the provider operators' subscriptions and csvs are set to be the dbaasplatform CR so that they will be deleted when uninstalling the dbaas operator (the owner reference of the dbaasplatform CR is the dbaas operator's csv).

It seems the [OwnerReferencesPermissionEnforcement](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement) admission controller is enabled for openshift by default. Because I got this error when testing it:

> 2021-11-18T21:52:04.109Z	ERROR	controller-runtime.manager.controller.dbaasplatform	Reconciler error	{"reconciler group": "dbaas.redhat.com", "reconciler kind": "DBaaSPlatform", "name": "dbaas-platform", "namespace": "openshift-dbaas-operator", "error": "clusterserviceversions.operators.coreos.com \"service-binding-operator.v0.10.0\" is forbidden: cannot set an ownerRef on a resource you can't delete: , <nil>"}

So I kept the `delete` permission to set the owner references, and also added `update` permission for `clusterserviceversions/finalizers` for the same reason.

tchughesiv: is that the route we decided to take here? to clean up other operator installs when dbaas is uninstalled? what about something like SBO?

xieshenzh: We try to remove all the platform resources that the dbaas operator is supposed to create, including the subscription and csv of the service binding operator. Especially with this PR, everything is created in the same namespace.

The whole feature of dbaas platform was brought from the managed kafka project. I don't think we fully discussed it. 
Personally I think all the platform resources should be independent of the dbaas operator. All the provider operators and dynamic plugins should not be created by the dbaas operator. But for now, the dbaas operator creates all these things and deletes as many as possible.

jianrongzhang89: @xieshenzh could you document with comments in the code, or in readme file that when dbaas operator is removed by OLM, what will be deleted and what will be left?

xieshenzh: @jianrongzhang89 documented with comments in the code: https://github.com/RHEcosystemAppEng/dbaas-operator/blob/5665e3efc2bfde909dbe299b4a2640aa4fd61f54/controllers/dbaasplatform_controller.go#L45-L65

tchughesiv: i see `Platform resources NOT to be removed or updated by the DBaaS operator`... so how do we handle updates to those components when necessary?

tchughesiv (commented):
added a question about UI & provider updates in future releases... otherwise lgt

tchughesiv (dismissed):

controllers/dbaasplatform_controller.go
xieshenzh: They will be updated, for example: https://github.com/RHEcosystemAppEng/dbaas-operator/blob/5665e3efc2bfde909dbe299b4a2640aa4fd61f54/controllers/reconcilers/console_plugin/plugin_installation_reconciler.go#L242

But we might not be able to support the upgrade for future releases, if we have to install these resources in a different way, for example using operators.

jianrongzhang89 (approved):

PR #72, opened 2021-11-23
DBAAS-201 DBaaS Operator error handling enhancements

jianrongzhang89:
## Description
This PR is to implement https://issues.redhat.com/browse/DBAAS-201 and enhance the reconciliation logic so that the operator sets the appropriate status for the DBaaSConnection and DBaaSInventory CRs. Typical cases include:
1) There is no tenant configured for the DBaaSInventory CR in the namespace
2) There is no valid provider found for the DBaaSInventory CR
3) The inventory for the DBaaSConnection CR does not exist
4) The provider of the inventory for the DBaaSConnection CR is not valid
 5) The inventory for the DBaaSConnection CR is not ready, ie, service discovery has not been completed successfully

In this PR, a new "Ready" condition type is added to DBaaSInventory and DBaaSConnection status conditions along with existing conditions synced from provider operators.

There are two changes to the reconciliation flows:
1) For DBaaSConnection:  check if the inventory is ready before further reconciliation (creating provider CR etc) is performed;
2) For DBaaSInvetory:  the function to retrieve the inventory was moved in front of the dbaastenant checking. This allows us to set a corresponding status condition in the CR in case there is no tenant configured for the target namespace.

## Type of change

- [x] Code Review completed
- [x] Verified independently by reviewer

Verification steps:
1) Create an OpenShift 4.9 RHDMS testing cluster and deploy the DBaaS Operator under openshift-dbaas-operator project.
2) Download attached zip file and unzip.
[error-handling-test.zip](https://github.com/RHEcosystemAppEng/dbaas-operator/files/7612534/error-handling-test.zip)
3) log in to the cluster as cluster admin.
4) oc new-project jianrong
Note: The default dbaas tenant is only configured for openshift-dbaas-operator
5) Create an inventory that does not have a provider registered: 
oc apply -f dbassinventory-notenant.yaml
dbaasinventory.dbaas.redhat.com/dbaas-jianrzha-notenant created
Check the created CR status. Example:
Status:
  Conditions:
    Last Transition Time:  2021-11-24T02:12:38Z
    Message:               failed to find DBaaS tenants
    Reason:                DBaaSTenantNotFound
    Status:                False
    Type:                  Ready
6) Now create a connection with above inventory:
 oc apply -f dbassconnection_notenant.yaml
dbaasconnection.dbaas.redhat.com/test-dbaas-connection-notenant created
Status:
  Conditions:
    Last Transition Time:  2021-11-26T01:38:47Z
    Message:               Inventory discovery not done
    Reason:                DBaaSInventoryNotReady
    Status:                False
    Type:                  Ready

7) Now switch to openshift-dbaas-operator where the default tenant is applied.
oc project openshift-dbaas-operator
8) oc apply -f dbassinventory_noprovider.yaml
dbaasinventory.dbaas.redhat.com/dbaas-jianrzha-no-provider created
Check the created CR status. Example:
Status:
  Conditions:
    Last Transition Time:  2021-11-24T02:15:41Z
    Message:               DBaaSProvider.dbaas.redhat.com "mongodb-atlas-registration-nonexist" not found
    Reason:                DBaaSProviderNotFound
    Status:                False
    Type:                  Ready
9) Now create a connection with previous connection:
oc apply -f dbassconnection_noprovider.yaml
dbaasconnection.dbaas.redhat.com/test-dbaas-connection-no-provider created
Check CR status:
Status:
  Conditions:
    Last Transition Time:  2021-11-26T01:41:50Z
    Message:               DBaaSProvider.dbaas.redhat.com "mongodb-atlas-registration-nonexist" not found
    Reason:                DBaaSProviderNotFound
    Status:                False
    Type:                  Ready

10) oc apply -f dbassinventory.yaml
dbaasinventory.dbaas.redhat.com/dbaas-jianrzha created
Check inventory CR status - note that its status is False.
Status:
  Conditions:
    Last Transition Time:  2021-11-26T02:01:29Z
    Message:               Secret "my-atlas-key" not found
    Reason:                InputError
    Status:                False
    Type:                  SpecSynced
    Last Transition Time:  2021-11-26T02:01:29Z
    Message:               DBaaS Provider Custom Resource created or updated
    Reason:                ProviderReconcileInprogress
    Status:                False
    Type:                  Ready
11) Now create a connection:
 oc apply -f dbassconnection.yaml
dbaasconnection.dbaas.redhat.com/test-dbaas-connection created
Check the connection status:
Status:
  Conditions:
    Last Transition Time:  2021-11-26T02:03:03Z
    Message:               Inventory discovery not done
    Reason:                DBaaSInventoryNotReady
    Status:                False
    Type:                  Ready
12) Create a file atlas-key.yaml as follows
apiVersion: v1
kind: Secret
metadata:
  name: my-atlas-key
type: Opaque
data:
  orgId: <your orgId>
  privateApiKey: <Your private API Key for Atlas>
  publicApiKey: <Your public API Key for Atlas>
13) Create the secret:
oc apply -f <Your private API Key>
Now check the above inventory status:
Status:
  Conditions:
    Last Transition Time:  2021-11-26T02:01:29Z
    Message:               GET https://cloud.mongodb.com/api/atlas/v1.0/groups: 403 (request "IP_ADDRESS_NOT_ON_ACCESS_LIST") IP address 18.189.136.122 is not allowed to access this resource.
    Reason:                BackendError
    Status:                False
    Type:                  SpecSynced
    Last Transition Time:  2021-11-26T02:01:29Z
    Message:               DBaaS Provider Custom Resource created or updated
    Reason:                ProviderReconcileInprogress
    Status:                False
    Type:                  Ready
and check the above connections sttaus:
Status:
  Conditions:
    Last Transition Time:  2021-11-26T02:03:03Z
    Message:               Inventory discovery not done
    Reason:                DBaaSInventoryNotReady
    Status:                False
    Type:                  Ready

15) Now we add the IP address (18.189.136.122 herein) into the IP access list in Atlas website.
Check the inventory status again - now you can see that the inventory has been successfully reconciled:
Status:
  Conditions:
    Last Transition Time:  2021-11-26T02:08:51Z
    Message:               Spec sync OK
    Reason:                SyncOK
    Status:                True
    Type:                  SpecSynced
    Last Transition Time:  2021-11-26T02:08:51Z
    Message:               Provider Custom Resource status sync completed
    Reason:                Ready
    Status:                True
    Type:                  Ready
  Instances:
    Instance ID:  608df625aa94426b4167e45f
    Instance Info:
      Connection Strings Standard Srv:  mongodb+srv://dbaas-cluster1.h38mo.mongodb.net
      Instance Size Name:               M0
      Project ID:                       608df5e652e1944293e8584a
      Project Name:                     Project 1
      Provider Name:                    TENANT
      Region Name:                      US_EAST_1
    Name:                               DBaaS-Cluster1
    Instance ID:                        60807282b4ab8d3b3c84be53
    Instance Info:
      Connection Strings Standard Srv:  mongodb+srv://test.iojsa.mongodb.net
      Instance Size Name:               M10
      Project ID:                       6065e15b16c0731bf3a2aefa
      Project Name:                     Project 2
      Provider Name:                    AWS
      Region Name:                      US_EAST_1
    Name:                               test
    Instance ID:                        606b2ffbc9a90e310e642482
    Instance Info:
      Connection Strings Standard Srv:  mongodb+srv://dbcreatedinatalas.iojsa.mongodb.net
      Instance Size Name:               M0
      Project ID:                       6065e15b16c0731bf3a2aefa
      Project Name:                     Project 2
      Provider Name:                    TENANT
      Region Name:                      US_EAST_1
    Name:                               DBCreatedInAtalas
    Instance ID:                        60b7a72f4877d05880c487d2
    Instance Info:
      Connection Strings Standard Srv:  mongodb+srv://test.fbbz5.mongodb.net
      Instance Size Name:               M10
      Project ID:                       60b798fea37f9f09acc1a154
      Project Name:                     jianrzha
      Provider Name:                    AWS
      Region Name:                      US_EAST_1
    Name:                               test

Now lets check the connection CR status:
status:
  conditions:
  - lastTransitionTime: "2021-11-26T02:11:15Z"
    message: ""
    reason: Ready
    status: "True"
    type: ReadyForBinding
  - lastTransitionTime: "2021-11-26T02:49:49Z"
    message: Provider Custom Resource status sync completed
    reason: Ready
    status: "True"
    type: Ready
  connectionInfoRef:
    name: atlas-connection-cm-c9f67
  credentialsRef:
    name: atlas-db-user-rhjjf

### DIFF
--- a/api/v1alpha1/dbaasprovider.go
+++ b/api/v1alpha1/dbaasprovider.go
@@ -21,6 +21,30 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// DBaaS condition types
+	DBaaSInventoryReadyType         string = "InventoryReady"
+	DBaaSInventoryProviderSyncType  string = "SpecSynced"
+	DBaaSConnectionProviderSyncType string = "ReadyForBinding"
+	DBaaSConnectionReadyType        string = "ConnectionReady"
+
+	// DBaaS condition reasons
+	Ready                       string = "Ready"
+	DBaaSTenantNotFound         string = "DBaaSTenantNotFound"
+	DBaaSProviderNotFound       string = "DBaaSProviderNotFound"
+	DBaaSInventoryNotFound      string = "DBaaSInventoryNotFound"
+	DBaaSInventoryNotReady      string = "DBaaSInventoryNotReady"
+	ProviderReconcileInprogress string = "ProviderReconcileInprogress"
+	ProviderParsingError        string = "ProviderParsingError"
+
+	// DBaaS condition messages
+	MsgProviderCRStatusSyncDone      string = "Provider Custom Resource status sync completed"
+	MsgProviderCRCreatedOrUpdated    string = "DBaaS Provider Custom Resource created or updated"
+	MsgProviderCRReconcileInProgress string = "DBaaS Provider Custom Resource reconciliation in progress"
+	MsgInventoryNotReady             string = "Inventory discovery not done"
+	MsgTenantNotFound                string = "Failed to find DBaaS tenants"
+)
+
 // DBaaSProviderSpec defines the desired state of DBaaSProvider
 type DBaaSProviderSpec struct {
 	// Provider contains information about database provider & platform
--- a/controllers/dbaasconnection_controller.go
+++ b/controllers/dbaasconnection_controller.go
@@ -18,10 +18,12 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
 	appv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/pointer"
@@ -47,7 +49,7 @@ type DBaaSConnectionReconciler struct {
 //
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
-func (r *DBaaSConnectionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *DBaaSConnectionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, recErr error) {
 	logger := ctrl.LoggerFrom(ctx, "DBaaS Connection", req.NamespacedName)
 
 	var connection v1alpha1.DBaaSConnection
@@ -55,72 +57,107 @@ func (r *DBaaSConnectionReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		if errors.IsNotFound(err) {
 			// CR deleted since request queued, child objects getting GC'd, no requeue
 			logger.Info("DBaaS Connection resource not found, has been deleted")
-			return ctrl.Result{}, nil
+			result, recErr = ctrl.Result{}, nil
+			return
 		}
 		logger.Error(err, "Error fetching DBaaS Connection for reconcile")
-		return ctrl.Result{}, err
+		result, recErr = ctrl.Result{}, err
+		return
 	}
 
-	if result, err := r.reconcileDevTopologyResource(&connection, ctx); err != nil {
+	var dbaasCond metav1.Condition
+	// This update will make sure the status is always updated in case of any errors or successful result
+	defer func(conn *v1alpha1.DBaaSConnection, cond *metav1.Condition) {
+		apimeta.SetStatusCondition(&conn.Status.Conditions, *cond)
+		if err := r.Client.Status().Update(ctx, conn); err != nil {
+			if errors.IsConflict(err) {
+				logger.V(1).Info("Connection modified, retry syncing spec")
+				// Re-queue and preserve existing recErr
+				result = ctrl.Result{Requeue: true}
+				return
+			}
+			logger.Error(err, "Could not update connection status")
+			if recErr == nil {
+				// There is no existing recErr. Set it to the status update error
+				recErr = err
+			}
+		}
+	}(&connection, &dbaasCond)
+
+	if res, err := r.reconcileDevTopologyResource(&connection, ctx); err != nil {
 		if errors.IsConflict(err) {
 			logger.Info("Deployment for Developer Topology view modified, retry reconciling")
-			return ctrl.Result{Requeue: true}, nil
+			result, recErr = ctrl.Result{Requeue: true}, nil
+			return
 		}
 		logger.Error(err, "Error reconciling Deployment for Developer Topology view")
-		return ctrl.Result{}, err
+		result, recErr = ctrl.Result{}, err
+		return
 	} else {
-		logger.Info("Deployment for Developer Topology view reconciled", "result", result)
+		logger.Info("Deployment for Developer Topology view reconciled", "result", res)
 	}
 
 	var inventory v1alpha1.DBaaSInventory
 	if err := r.Get(ctx, types.NamespacedName{Namespace: connection.Spec.InventoryRef.Namespace, Name: connection.Spec.InventoryRef.Name}, &inventory); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Error(err, "DBaaS Inventory resource not found", "DBaaS Inventory", connection.Spec.InventoryRef)
+			dbaasCond = metav1.Condition{Type: v1alpha1.DBaaSConnectionReadyType, Status: metav1.ConditionFalse, Reason: v1alpha1.DBaaSInventoryNotFound, Message: err.Error()}
+			result, recErr = ctrl.Result{}, err
+			return
+		}
 		logger.Error(err, "Error fetching DBaaS Inventory resource reference for DBaaS Connection", "DBaaS Inventory", connection.Spec.InventoryRef)
-		return ctrl.Result{}, err
+		result, recErr = ctrl.Result{}, err
+		return
 	}
 
 	provider, err := r.getDBaaSProvider(inventory.Spec.ProviderRef.Name, ctx)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			logger.Error(err, "Requested DBaaS Provider is not configured in this environment", "DBaaS Provider", inventory.Spec.ProviderRef)
-			return ctrl.Result{}, err
+			dbaasCond = metav1.Condition{Type: v1alpha1.DBaaSConnectionReadyType, Status: metav1.ConditionFalse, Reason: v1alpha1.DBaaSProviderNotFound, Message: err.Error()}
+			result, recErr = ctrl.Result{}, err
+			return
 		}
 		logger.Error(err, "Error reading configured DBaaS Provider", "DBaaS Provider", inventory.Spec.ProviderRef)
-		return ctrl.Result{}, err
+		result, recErr = ctrl.Result{}, err
+		return
 	}
 	logger.Info("Found DBaaS Provider", "DBaaS Provider", inventory.Spec.ProviderRef)
 
+	// The inventory must be in ready status before we can move on
+	invCond := apimeta.FindStatusCondition(inventory.Status.Conditions, v1alpha1.DBaaSInventoryReadyType)
+	if invCond == nil || invCond.Status == metav1.ConditionFalse {
+		err := fmt.Errorf(v1alpha1.MsgInventoryNotReady)
+		logger.Error(err, "Inventory is not ready", "Inventory", inventory.Name, "Namespace", inventory.Namespace)
+		dbaasCond = metav1.Condition{Type: v1alpha1.DBaaSConnectionReadyType, Status: metav1.ConditionFalse, Reason: v1alpha1.DBaaSInventoryNotReady, Message: v1alpha1.MsgInventoryNotReady}
+		result, recErr = ctrl.Result{}, err
+		return
+	}
+
 	providerConnection := r.createProviderObject(&connection, provider.Spec.ConnectionKind)
-	if result, err := r.reconcileProviderObject(providerConnection, r.providerObjectMutateFn(&connection, providerConnection, connection.Spec.DeepCopy()), ctx); err != nil {
+	if res, err := r.reconcileProviderObject(providerConnection, r.providerObjectMutateFn(&connection, providerConnection, connection.Spec.DeepCopy()), ctx); err != nil {
 		if errors.IsConflict(err) {
 			logger.Info("Provider Connection modified, retry syncing spec")
-			return ctrl.Result{Requeue: true}, nil
+			result, recErr = ctrl.Result{Requeue: true}, nil
+			return
 		}
 		logger.Error(err, "Error reconciling Provider Connection resource")
-		return ctrl.Result{}, err
+		result, recErr = ctrl.Result{}, err
+		return
 	} else {
-		logger.Info("Provider Connection resource reconciled", "result", result)
+		logger.Info("Provider Connection resource reconciled", "result", res)
 	}
 
 	var DBaaSProviderConnection v1alpha1.DBaaSProviderConnection
 	if err := r.parseProviderObject(providerConnection, &DBaaSProviderConnection); err != nil {
 		logger.Error(err, "Error parsing the Provider Connection resource")
-		return ctrl.Result{}, err
+		dbaasCond = metav1.Condition{Type: v1alpha1.DBaaSConnectionReadyType, Status: metav1.ConditionFalse, Reason: v1alpha1.ProviderParsingError, Message: err.Error()}
+		result, recErr = ctrl.Result{}, err
+		return
 	}
-	if err := r.reconcileDBaaSObjectStatus(&connection, func() error {
-		DBaaSProviderConnection.Status.DeepCopyInto(&connection.Status)
-		return nil
-	}, ctx); err != nil {
-		if errors.IsConflict(err) {
-			logger.Info("DBaaS Connection modified, retry syncing status")
-			return ctrl.Result{Requeue: true}, nil
-		}
-		logger.Error(err, "Error updating the DBaaS Connection status")
-		return ctrl.Result{}, err
-	} else {
-		logger.Info("DBaaS Connection status updated")
-	}
-
-	return ctrl.Result{}, nil
+	dbaasCond = *mergeConnectionStatus(&connection, &DBaaSProviderConnection)
+	result, recErr = ctrl.Result{}, nil
+	return
 }
 
 // SetupWithManager sets up the controller with the Manager.
@@ -180,4 +217,20 @@ func (r *DBaaSConnectionReconciler) deploymentMutateFn(connection *v1alpha1.DBaa
 		}
 		return nil
 	}
+}
+
+// mergeConnectionStatus: merge the status from DBaaSProviderConnection into the current DBaaSConnection status
+func mergeConnectionStatus(conn *v1alpha1.DBaaSConnection, providerConn *v1alpha1.DBaaSProviderConnection) *metav1.Condition {
+
+	cond := apimeta.FindStatusCondition(conn.Status.Conditions, v1alpha1.DBaaSConnectionReadyType)
+	providerConn.Status.DeepCopyInto(&conn.Status)
+	if cond != nil {
+		conn.Status.Conditions = append(conn.Status.Conditions, *cond)
+	}
+	// Update connection status condition (type: DBaaSConnectionReadyType) based on the provider status
+	specSync := apimeta.FindStatusCondition(providerConn.Status.Conditions, v1alpha1.DBaaSConnectionProviderSyncType)
+	if cond != nil && specSync != nil && specSync.Status == metav1.ConditionTrue {
+		return &metav1.Condition{Type: v1alpha1.DBaaSConnectionReadyType, Status: metav1.ConditionTrue, Reason: v1alpha1.Ready, Message: v1alpha1.MsgProviderCRStatusSyncDone}
+	}
+	return &metav1.Condition{Type: v1alpha1.DBaaSConnectionReadyType, Status: metav1.ConditionFalse, Reason: v1alpha1.ProviderReconcileInprogress, Message: v1alpha1.MsgProviderCRReconcileInProgress}
 }
--- a/controllers/dbaasinventory_controller.go
+++ b/controllers/dbaasinventory_controller.go
@@ -21,6 +21,8 @@ import (
 
 	"github.com/RHEcosystemAppEng/dbaas-operator/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 )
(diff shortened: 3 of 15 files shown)


api/v1alpha1/dbaasprovider.go
```diff
@@ -21,6 +21,31 @@ import (
...
+// DBaaS condition types
+const (
+	DBaaSInventoryReadyType         string = "Ready"
+	DBaaSInventoryProviderSyncType  string = "SpecSynced"
+	DBaaSConnectionProviderSyncType string = "ReadyForBinding"
+	DBaaSConnectionReadyType        string = "Ready"
+)
+
+// DBaaS condition reasons
+const (
```
redhatHameed, line 33: can make one const var? and add the comments for each section

controllers/dbaasconnection_controller.go
```diff
@@ -181,3 +207,64 @@ func (r *DBaaSConnectionReconciler) deploymentMutateFn(connection *v1alpha1.DBaa
 		return nil
 	}
 }
+
+// SetConnectionCondition sets a condition on the status object. If the condition already
+// exists, it will be replaced. SetCondition does not update the resource in
+// the cluster.
+func SetConnectionCondition(conn *v1alpha1.DBaaSConnection, condType string, status metav1.ConditionStatus, reason, msg string) {
```
redhatHameed, line 214: is it possible to use a builtin function from ->  k8s.io/apimachinery/pkg/apis/meta/v1 https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go#L30
done in crunchy bridge operator 
https://github.com/CrunchyData/crunchy-bridge-operator/blob/master/controllers/dbaas.redhat.com/conditions.go#L44

controllers/dbaasplatform_controller.go
```diff
@@ -242,21 +243,21 @@ func (r *DBaaSPlatformReconciler) createPlatformCR(ctx context.Context, serverCl
 func (r *DBaaSPlatformReconciler) getInstallationPlatforms() []dbaasv1alpha1.PlatformsName {
 
 	return []dbaasv1alpha1.PlatformsName{
-		dbaasv1alpha1.CrunchyBridgeInstallation,
-		dbaasv1alpha1.MongoDBAtlasInstallation,
 		dbaasv1alpha1.DBassDynamicPluginInstallation,
 		dbaasv1alpha1.ConsoleTelemetryPluginInstallation,
+		dbaasv1alpha1.CrunchyBridgeInstallation,
+		dbaasv1alpha1.MongoDBAtlasInstallation,
 		dbaasv1alpha1.ServiceBindingInstallation,
```
redhatHameed, line 250: +1 changing the order of platform installation plugin first then providers operators

redhatHameed (commented):
Thanks added a few comments

controllers/dbaasinventory_controller.go
```diff
@@ -46,94 +48,109 @@ type DBaaSInventoryReconciler struct {
...
-		//
-		// Inventory RBAC
-		//
-		if err := r.reconcileInventoryRbacObjs(ctx, inventory, tenantList); err != nil {
-			return ctrl.Result{}, err
-		}
+	if len(tenantList.Items) == 0 {
+		logger.Error(err, "No DBaaS tenant found for the target namespace", "Namespace", req.Namespace)
+		SetInventoryCondition(&inventory, v1alpha1.DBaaSInventoryReadyType, metav1.ConditionFalse, v1alpha1.DBaaSTenantNotFound, "failed to find DBaaS tenants")
+		return ctrl.Result{}, fmt.Errorf(v1alpha1.DBaaSTenantNotFound)
```
tchughesiv, line 79: not sure we want to throw errors during inventory reconcile for every inventory that doesn't reside in a tenant namespace. this could be a common occurrence in certain clusters and isn't necessarily something to be concerned about.

If we think we want to display this, should it instead be a debug message? warning?

controllers/dbaasinventory_controller.go
```diff
@@ -46,94 +48,109 @@ type DBaaSInventoryReconciler struct {
...
+		return ctrl.Result{}, err
+	}
+
+	// This update will make sure the status is always updated in case of any errors or successful result
+	defer func(inv *v1alpha1.DBaaSInventory) {
+		err := r.Client.Status().Update(ctx, inv)
+		if err != nil {
+			logger.Info(fmt.Sprintf("Could not update resource status:%v", err))
+		}
+	}(&inventory)
```
tchughesiv, line 68: This PR has to do with error handling, bit these are significant changes likely to negatively impact reconcile flow. This block of code should stay where it was. The reason is that we don't want to even attempt a reconciliation of an inventory unless it exists in a "tenant namespace". We should just ignore all other inventories in the cluster.

tchughesiv (changes requested):

controllers/dbaasinventory_controller.go
tchughesiv: looking at this closer, i agree it makes sense... as long as we remove the below error.

controllers/dbaasinventory_controller.go
jianrongzhang89: Yes, you are right. They are not really errors and will clog the pod logs. I think changing to debug is good.

@jeremyary @xieshenzh @redhatHameed do you think debug message is good here, or should a warning be logged instead?

controllers/dbaasconnection_controller.go
jianrongzhang89: Yes this is a good idea. I changed to use the standard function call for both setcondition and getcondition calls and upated the PR.

controllers/dbaasconnection_controller.go
```diff
@@ -98,16 +123,20 @@ func (r *DBaaSConnectionReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 		logger.Info("Provider Connection resource reconciled", "result", result)
 	}
 
 	var DBaaSProviderConnection v1alpha1.DBaaSProviderConnection
 	if err := r.parseProviderObject(providerConnection, &DBaaSProviderConnection); err != nil {
 		logger.Error(err, "Error parsing the Provider Connection resource")
+		setConnectionCondition(&connection, v1alpha1.DBaaSConnectionReadyType, metav1.ConditionFalse, v1alpha1.ProviderParsingError, err.Error())
 		return ctrl.Result{}, err
 	}
 	if err := r.reconcileDBaaSObjectStatus(&connection, func() error {
```
xieshenzh, line 138: Since the status is updated in the deferred function, it is not necessary to update it here.
You may use the function `reconcileDBaaSObjectStatus` as the deferred function, or delete it because it is not used anywhere else.
And make sure check if the err is `Conflict`, we don't have to write `Conflict` error in the logs.

controllers/dbaasinventory_controller.go
redhatHameed: How about adding the info message? if not then you can make it debug.

controllers/dbaasconnection_controller.go
jianrongzhang89: Thanks @xieshenzh, yes this can be simplified. I removed reconcileDBaaSObjectStatus all together and updated the PR accordingly.

controllers/dbaasinventory_controller.go
jianrongzhang89: I used logger.V(1).Info() for this message.

api/v1alpha1/dbaasprovider.go
```diff
@@ -21,6 +21,27 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// DBaaS condition types
+	DBaaSInventoryReadyType         string = "Ready"
```
redhatHameed, line 26: @jianrongzhang89 why we need `Type: Ready` I mean why not use the same type `ReadyForBinding`or `SpecSynced`

jianrongzhang89: @redhatHameed we need to use a different condition type to separate dbaas specific condition from ReadyForBinding or SpecSynced which is copied from provider provider operator CRs.

tchughesiv (dismissed):

controllers/dbaasinventory_controller.go
```diff
@@ -200,3 +206,41 @@ func inventoryRbacObjs(inventory v1alpha1.DBaaSInventory, tenantList v1alpha1.DB
...
+func getInventoryCondition(inv *v1alpha1.DBaaSInventory, condType string) *metav1.Condition {
+	return apimeta.FindStatusCondition(inv.Status.Conditions, condType)
+}
+
+// mergeInventoryStatus: merge the status from DBaaSProviderInventory into the current DBaaSInventory status
+func mergeInventoryStatus(inv *v1alpha1.DBaaSInventory, providerInv *v1alpha1.DBaaSProviderInventory) error {
+	status := providerInv.Status.DeepCopy()
+	cond := getInventoryCondition(inv, v1alpha1.DBaaSInventoryReadyType)
+	if cond != nil {
+		status.Conditions = append(status.Conditions, *cond)
```
xieshenzh, line 234: It seems the condition with type `v1alpha1.DBaaSInventoryReadyType` will always be reset at line 240 or 244. In this case, it doesn't make sense to set it here.

xieshenzh (changes requested):

api/v1alpha1/dbaasprovider.go
redhatHameed: can we change the name of type from 'Ready' to 'ReconcileSuccess' WDYT?

controllers/dbaasconnection_controller.go
```diff
@@ -98,29 +123,20 @@ func (r *DBaaSConnectionReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		logger.Error(err, "Error reconciling Provider Connection resource")
 		return ctrl.Result{}, err
 	} else {
+		if result == controllerutil.OperationResultCreated || result == controllerutil.OperationResultUpdated {
+			setConnectionCondition(&connection, v1alpha1.DBaaSConnectionReadyType, metav1.ConditionFalse, v1alpha1.ProviderReconcileInprogress, "DBaaS Provider Custom Resource created or updated")
```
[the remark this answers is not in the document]
jianrongzhang89: Sounds good. Done

controllers/dbaasconnection_controller.go
```diff
@@ -181,3 +197,41 @@ func (r *DBaaSConnectionReconciler) deploymentMutateFn(connection *v1alpha1.DBaa
...
+		status.Conditions = append(status.Conditions, *cond)
+	}
+	status.DeepCopyInto(&conn.Status)
+	// Update connection status condition (type: DBaaSConnectionReadyType) based on the provider status
+	specSync := apimeta.FindStatusCondition(providerConn.Status.Conditions, v1alpha1.DBaaSConnectionProviderSyncType)
+	if cond != nil && specSync != nil && specSync.Status == metav1.ConditionTrue {
+		setConnectionCondition(conn, v1alpha1.DBaaSConnectionReadyType, metav1.ConditionTrue, v1alpha1.Ready, v1alpha1.MsgProviderStatusSyncDone)
+		return nil
+	}
+	msg := "DBaaS Provider Custom Resource reconcilation in progress"
```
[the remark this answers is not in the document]
jianrongzhang89: Corrected and defined as a const.

controllers/dbaasconnection_controller.go
```diff
@@ -181,3 +197,41 @@ func (r *DBaaSConnectionReconciler) deploymentMutateFn(connection *v1alpha1.DBaa
...
+		Message: msg,
+	}
+	apimeta.SetStatusCondition(&conn.Status.Conditions, newCondition)
+}
+
+// getConnectionCondition return the condition with the passed condition type from
+// the status object. If the condition is not already present, return nil
+func getConnectionCondition(conn *v1alpha1.DBaaSConnection, condType string) *metav1.Condition {
+	return apimeta.FindStatusCondition(conn.Status.Conditions, condType)
+}
```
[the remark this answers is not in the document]
jianrongzhang89: The set function is a useful help function so I kept it but  I combined the setConnectionCondition and setInventoryCOndition into one and moved it to api package. The getConnectionCondition is replaced with apimeta function calls directly.

controllers/dbaasinventory_controller.go
```diff
@@ -200,3 +206,41 @@ func inventoryRbacObjs(inventory v1alpha1.DBaaSInventory, tenantList v1alpha1.DB
...
+}
+
+// getInventoryCondition return the condition with the passed condition type from
+// the status object. If the condition is not already present, return nil
+func getInventoryCondition(inv *v1alpha1.DBaaSInventory, condType string) *metav1.Condition {
+	return apimeta.FindStatusCondition(inv.Status.Conditions, condType)
+}
+
+// mergeInventoryStatus: merge the status from DBaaSProviderInventory into the current DBaaSInventory status
+func mergeInventoryStatus(inv *v1alpha1.DBaaSInventory, providerInv *v1alpha1.DBaaSProviderInventory) error {
```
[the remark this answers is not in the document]
jianrongzhang89: Actually, because DBaaSInventoryStatus and DBaaSConnectionStatus are two different data structures, mergeInventoryStatus and mergeConnectionStatus can not be easily refactored into one so I prefer to keep them separate.

controllers/dbaasinventory_controller.go
jianrongzhang89: Actually, only the status, reason and msg are reset again later,  lastTransitionTime does not always gets reset depending on the specific values.

api/v1alpha1/dbaasprovider.go
jianrongzhang89: This could change.  My preference is the short version of "Ready" though. @tchughesiv @xieshenzh and @jeremyary what do you think?

controllers/dbaasinventory_controller.go
xieshenzh: If the condition is never used (i.e. its status, reason and msg will always be reset), I don't think there is any significance of its old value of lastTransitionTime. 
And it seems any condition of type `v1alpha1.DBaaSInventoryReadyType` is created by this controller, so I don't think it is a problem if this controller sets lastTransitionTime again in this function.

xieshenzh: Actually, I don't think the `Reconcile` function should set any condition of type `v1alpha1.DBaaSInventoryReadyType` unless it returns with requeue or error. Because the value will never be used, it will always be reset in this function.
On the other hand, if the `Reconcile` function doesn't set any condition of type `v1alpha1.DBaaSInventoryReadyType`, there is no need to read it in this function.

controllers/dbaasinventory_controller.go
```diff
@@ -200,3 +207,21 @@ func inventoryRbacObjs(inventory v1alpha1.DBaaSInventory, tenantList v1alpha1.DB
...
+	}
+	status.DeepCopyInto(&inv.Status)
+	// Update inventory status condition (type: DBaaSInventoryReadyType) based on the provider status
+	specSync := apimeta.FindStatusCondition(providerInv.Status.Conditions, v1alpha1.DBaaSInventoryProviderSyncType)
+	if cond != nil && specSync != nil && specSync.Status == metav1.ConditionTrue {
+		v1alpha1.SetStatusCondition(&inv.Status.Conditions, v1alpha1.DBaaSInventoryReadyType, metav1.ConditionTrue, v1alpha1.Ready, v1alpha1.MsgProviderCRStatusSyncDone)
+		return nil
+	}
+	v1alpha1.SetStatusCondition(&inv.Status.Conditions, v1alpha1.DBaaSInventoryReadyType, metav1.ConditionFalse, v1alpha1.ProviderReconcileInprogress, v1alpha1.MsgProviderCRReconcileInProgress)
+	return fmt.Errorf(v1alpha1.MsgProviderCRReconcileInProgress)
```
xieshenzh, line 226: This function should not return error here. `ProviderCRReconcileInProgress` is not an error. We don't have to requeue the Reconcile either. Once the provider finishes reconcile, it should update the provider CR which will trigger this Reconcile.

jianrongzhang89: Agreed. Code updated.

controllers/dbaasinventory_controller.go
jianrongzhang89: @xieshenzh thank you very much for your suggestions! Based on our discussions, I simplified the code. Please review again.

controllers/dbaasconnection_controller.go
```diff
@@ -61,6 +63,16 @@ func (r *DBaaSConnectionReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, err
 	}
 
+	var dbaasCond metav1.Condition
+	// This update will make sure the status is always updated in case of any errors or successful result
+	defer func(conn *v1alpha1.DBaaSConnection, dbaasCond *metav1.Condition) {
+		apimeta.SetStatusCondition(&conn.Status.Conditions, *dbaasCond)
+		err := r.Client.Status().Update(ctx, conn)
+		if err != nil {
```
xieshenzh, line 71: I think we can check if the error `IsConflict` here. `Reconcile` should be requeued if there is conflict when updating the status of the CR.
And we should alter the return values of function `Reconcile`, if we want to requeue or there is other error for updating the status of the CR.

controllers/dbaasconnection_controller.go
```diff
@@ -74,6 +86,11 @@ func (r *DBaaSConnectionReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	var inventory v1alpha1.DBaaSInventory
 	if err := r.Get(ctx, types.NamespacedName{Namespace: connection.Spec.InventoryRef.Namespace, Name: connection.Spec.InventoryRef.Name}, &inventory); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Error(err, "DBaaS Inventory resource not found", "DBaaS Inventory", connection.Spec.InventoryRef)
+			dbaasCond = metav1.Condition{Type: v1alpha1.DBaaSConnectionReadyType, Status: metav1.ConditionFalse, Reason: v1alpha1.DBaaSInventoryNotFound, Message: err.Error()}
+			return ctrl.Result{}, err
+		}
 		logger.Error(err, "Error fetching DBaaS Inventory resource reference for DBaaS Connection", "DBaaS Inventory", connection.Spec.InventoryRef)
```
xieshenzh, line 94: Should we also update the `DBaaSInventoryReadyType` condition for other errors like the one in this line? 
I am just curious why the condition is only updated for some of the error cases.

controllers/dbaasconnection_controller.go
jianrongzhang89: Agreed. Code updated to use named return values.

controllers/dbaasconnection_controller.go
jianrongzhang89: This intentional. I tried to separate the "business logic" errors (for instance tenant not found) from temporary errors (such us errors in retrieving an object from k8s) and the latter can normally get resolved through future reconciliations.

controllers/dbaasconnection_controller.go
xieshenzh: Thanks. In the case that there are already errors or requeue request before updating the status, I don't think we should overwrite the errors or requeue request with the ones caused by updating status.
So I suggest we check if the return values are already set before altering the return values, when there is error for updating the status.

jianrongzhang89: Thanks! I updated the code as we discussed.

xieshenzh (dismissed):

api/v1alpha1/dbaasprovider.go
redhatHameed: @jianrongzhang89 do we need this change. ? or want to keep 'Ready'

jeremyary:
@jianrongzhang89 can we clean up conflicts here, please? After that, I'll get it merged up. Thanks!!

jeremyary:
no-op comment, bumping to trigger actions

api/v1alpha1/dbaasprovider.go
jianrongzhang89: @redhatHameed yes. Changed to "InventoryReady" and "ConnectionReady".

jianrongzhang89:
> @jianrongzhang89 can we clean up conflicts here, please? After that, I'll get it merged up. Thanks!!

@jeremyary @tchughesiv The PR is rebased and all conflicts resolved. I retested and all cases passed.

redhatHameed (dismissed):

controllers/dbaasinventory_controller.go
```diff
@@ -39,74 +41,96 @@ type DBaaSInventoryReconciler struct {
...
-				// CR deleted since request queued, child objects getting GC'd, no requeue
-				return ctrl.Result{}, nil
-			}
-			logger.Error(err, "Error fetching DBaaS Inventory for reconcile")
-			return ctrl.Result{}, err
-		}
+	if len(tenantList.Items) == 0 {
+		logger.Info("No DBaaS tenant found for the target namespace", "Namespace", req.Namespace)
+		dbaasCond = metav1.Condition{Type: v1alpha1.DBaaSInventoryReadyType, Status: metav1.ConditionFalse, Reason: v1alpha1.DBaaSTenantNotFound, Message: v1alpha1.MsgTenantNotFound}
+		result, recErr = ctrl.Result{Requeue: true}, nil
```
tchughesiv, line 88: why Requeue here?

tchughesiv (changes requested):

controllers/dbaasinventory_controller.go
jianrongzhang89: This is required so that the inventory gets reconciled again when the user adds a tenant.

tchughesiv: won't this cause an infinite loop though, if no tenant for that namespace exists? i believe the requeue is initiated immediately.

jianrongzhang89: You are right. It should not be requeued then. I just updated the PR.

tchughesiv (approved):

tchughesiv:
lgtm

jeremyary (approved):

PR #90, opened 2021-12-23
Fix memory issue when setting owner references to CSVs

xieshenzh:
## Description

For the cluster-scoped operators, every namespace has a copy of the CSVs.

So when running load test with thousands of namespaces, setting owner references to the CSVs of mongodb atlas operator, crunchy bridge operator and service binding operator will cause the dbaas operator to load the thousands copies of CSVs to the cache, as a result the memory limit is exceeded.

The solution is to bypass the cache and read the CSVs directly from the API server.

## Type of change

- [ ] Code Review completed


xieshenzh:
Fix for https://github.com/RHEcosystemAppEng/dbaas-operator/pull/84

tchughesiv:
do we know what the impact to the stability of the api server would be with a change like this? how often are these calls made?

xieshenzh:
> do we know what the impact to the stability of the api server would be with a change like this? how often are these calls made?

I could reproduce the memory issue by simply installing the dbaas operator. There is no stability impact by only installing the dbaas operator based on my observation. 
If we want to create thousands of namespaces at the same time, the impact of creating the namespaces and other objects is definitely much more significant than this. We will have to run load test to figure out the difference.

At most three separate calls are made to read a CSV by its name and namespace for reconciling the dbaasplatform CR once. 
There is only one instance of the dbaasplatform CR, and the reconciling takes place most often during the startup of the operator to install the platform resources, around three times or less in every one second in my test. So there could be around ten calls to the api server made in every one second.

bundle/manifests/dbaas-operator.clusterserviceversion.yaml
```diff
@@ -347,15 +347,31 @@ spec:
...
+          - update
+          - watch
+        - apiGroups:
+          - operators.coreos.com
+          resources:
+          - clusterserviceversions
           - subscriptions
           verbs:
           - create
+          - delete
```
tchughesiv, line 364: do we need create or delete perms for csv's? maybe split out csv perms to only what we need?

controllers/dbaasplatform_controller.go
```diff
@@ -242,53 +238,40 @@ func (r *DBaaSPlatformReconciler) createPlatformCR(ctx context.Context, serverCl
...
 }
-func (r *DBaaSPlatformReconciler) getCleanupPlatforms() []dbaasv1alpha1.PlatformsName {
-	// since the operator cannot clean up resources automatically
-	// disable the cleanup to reduce the permission the operator needs
 
-	// will consider changing the scope of the dbaasplatform CR to be cluster-scoped
-	// and use ownerreferences to support auto cleanup in the future
+// Cleanup will be done by setting the dbaasplatform CR as the owner references for the namespace-scoped resources in the same namespace
+// Cluster-scoped resources and resources in other namespaces will not be removed or updated
+func (r *DBaaSPlatformReconciler) getCleanupPlatforms() []dbaasv1alpha1.PlatformsName {
```
tchughesiv, line 251: is this function still necessary if just returning `nil`?

tchughesiv (changes requested):

controllers/reconcilers/csv/csv_reconciler.go
```diff
@@ -69,6 +74,35 @@ func (r *Reconciler) Cleanup(ctx context.Context, cr *alpha1.DBaaSPlatform) (alp
...
+	}
+	existingGV, err := schema.ParseGroupVersion(existing.APIVersion)
+	if err != nil {
+		return false, err
+	}
+	equal := refGV.Group == existingGV.Group && ref.Kind == existing.Kind && ref.Name == existing.Name
+	return equal, nil
+}
+
 func GetDBaaSOperatorCSV(namespace string, ctx context.Context, serverClient k8sclient.Client) (*v1alpha1.ClusterServiceVersion, error) {
```
tchughesiv, line 106: in this function, you're still doing a list of the csv's... which would force a cache of the csv's anyway. if bypassing cache for `get`, we should also do so for a csv `list`.

tchughesiv (changes requested):

controllers/dbaas_reconciler.go
```diff
@@ -27,12 +27,13 @@ var inventoryNamespaceKey = ".spec.inventoryNamespace"
 type DBaaSReconciler struct {
 	client.Client
 	*runtime.Scheme
+	client.Reader
```
tchughesiv, line 30: if, with this change, we're only ever retrieving CSV's from the install namespace... the memory footprint for said csv's should be minimal. Are we sure we need to bypass cache for a single namespace worth of csv's?

controllers/reconcilers/csv/csv_reconciler.go
xieshenzh: The `serverClient` doesn't use cache: https://github.com/RHEcosystemAppEng/dbaas-operator/blob/beb6d1e43f265bff9d38dc0c09ee63335b72bfb6/controllers/dbaasplatform_controller.go#L172

The code is executed before the controller starts.

controllers/dbaas_reconciler.go
xieshenzh: It seems it only supports restricting the manager's cache to watch all the namespace-scoped objects in the desired namespace, and we cannot do it only for csv: https://github.com/kubernetes-sigs/controller-runtime/blob/e52a8b1fba620f42bf6be7c6904c77a6da342898/pkg/manager/manager.go#L189-L195

xieshenzh: The operator only reads the csv when installing the platform resources. It is not supposed to happen very often.

tchughesiv: but we aren't watching csv's... only retrieving certain ones. right? seems odd it would still try to cache ALL csv's cluster-wide.

xieshenzh: You are right, and the 'namespace' settings are not actually for the client. 
But the default cache uses informers as well, which consumes resources: https://github.com/kubernetes-sigs/controller-runtime/blob/e52a8b1fba620f42bf6be7c6904c77a6da342898/pkg/cache/cache.go#L150

There is an option to bypass cache for all csv: https://github.com/kubernetes-sigs/controller-runtime/blob/e52a8b1fba620f42bf6be7c6904c77a6da342898/pkg/cluster/cluster.go#L119
We can use this option if we want to always bypass the cache when reading csv. For now, we only read csv when setting ownerreferences.

tchughesiv: I guess we should explore both? first, let's make sure moving to the "reader" interface has a significant impact memory usage at scale. second, test the cache disable option and see if that's a better alternative... might be a better way to go?

tchughesiv: might gut tells me `ClientDisableCacheFor` is the way to go here. simpler and ensures cache is bypassed.

controllers/dbaasplatform_controller.go
```diff
@@ -208,24 +206,19 @@ func (r *DBaaSPlatformReconciler) createPlatformCR(ctx context.Context, serverCl
...
 			},
 
 			Spec: dbaasv1alpha1.DBaaSPlatformSpec{
 				Name: "Database as a Service",
 			},
 		}
 
+		if owner, err := reconcilers.GetDBaaSOperatorCSV(namespace, ctx, serverClient); err != nil {
+			return nil, fmt.Errorf("could not create dbaas platform intallation CR: %w", err)
+		} else if err := ctrl.SetControllerReference(owner, cr, r.Scheme); err != nil {
```
jianrongzhang89, line 218: Here "else if" can be replaced with if to make the code easier to understand.

bundle/manifests/dbaas-operator.clusterserviceversion.yaml
tchughesiv: seems `create` verb was removed, but not `delete`

tchughesiv (changes requested):

xieshenzh:
DBaaS operator pod resource metrics for load test  with 2000 users, projects and tenants.
It does take very long time to install the platform resources (i.e. operators), though.
Screenshot from 2022-01-05 16-43-44

controllers/dbaasplatform_controller.go
xieshenzh: done

controllers/dbaas_reconciler.go
xieshenzh: done

tchughesiv (approved):

PR #95, opened 2022-01-12
performance tune controllers

tchughesiv:

## Description
in large perf test clusters with 4k tenants/namespaces... the initial reconciles, on operator startup, take well over an hour to complete. this means new tenant/inventory events can't be processed until this startup queue completes.

with this PR, we reduce that startup processing time to under 14 minutes. also, we reduce openshift api calls tenfold which benefits clusters of any size. max concurrency for inventory and connection reconciles is increased to 2.

## Type of change

- [x] Code Review completed
- [x] Verified independently by reviewer


controllers/dbaas_authorization.go
```diff
@@ -23,17 +23,128 @@ import (
...
+}
+
+func getRequest(nameNS string) reconcile.Request {
+	return reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      nameNS,
+			Namespace: nameNS,
+		},
+	}
+}
```
xieshenzh, line 143: I think the code would be clearer if these functions are moved to a separate `*_controller.go` file. The rest of the  `DBaaSAuthzReconciler` functions could be used as utilities and moved to a file like `dbaas_authz_reconciler.go` which is similar to `dbaas_reconciler.go`.

xieshenzh (approved):
It looks good to me. Just leave a minor cosmetic suggestion. Thanks.

controllers/dbaas_authorization.go
tchughesiv: done, kept authz interface functions in the `authz_controller.go` file. let me know if you still think it warrants a separate `authz_reconciler.go` file

controllers/dbaas_reconciler.go
```diff
@@ -167,6 +342,102 @@ func GetInstallNamespace() (string, error) {
...
+
+// returns a subset of the provided slices, after removing certain entries
+func removeFromSlice(entriesToRemove, fromSlice []string) []string {
+	r := []string{}
+	for _, val := range fromSlice {
+		if !contains(entriesToRemove, val) {
+			r = append(r, val)
+		}
+
+	}
```
jeremyary, line 418: it's times like these golang makes you feel weird for having to write this stuff lol

controllers/dbaasconnection_controller.go
```diff
@@ -168,6 +168,9 @@ func (r *DBaaSConnectionReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 func (r *DBaaSConnectionReconciler) SetupWithManager(mgr ctrl.Manager) (controller.Controller, error) {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.DBaaSConnection{}).
+		WithOptions(
+			controller.Options{MaxConcurrentReconciles: 2},
+		).
```
jeremyary, line 173: TIL another cool manager option, thanks. So much useful stuff it can do that's really not talked about that I've found.

jeremyary (approved):

controllers/dbaas_authorization.go
xieshenzh: Thanks. I mean `DBaaSAuthzReconciler` and `DBaaSTenantReconciler` are independent controllers. The reason why `DBaaSTenantReconciler` has a [reference to `DBaaSAuthzReconciler`](https://github.com/RHEcosystemAppEng/dbaas-operator/blob/75e23665b402201d173fba79b21391cb2829860c/controllers/dbaastenant_controller.go#L31) is to [invoke function `reconcileTenantAuthz`](https://github.com/RHEcosystemAppEng/dbaas-operator/blob/75e23665b402201d173fba79b21391cb2829860c/controllers/dbaastenant_controller.go#L57) which is actually only called in this scenario.

We could use a reference to `AuthorizationV1Client` directly in `DBaaSTenantReconciler` and move the related functions  to `DBaaSTenantReconciler`, instead of embedding the `DBaaSAuthzReconciler` controller to `DBaaSTenantReconciler`. 
Or we could extract the utility functions from `DBaaSAuthzReconciler` to a new type, just like what we did with `dbaas_reconciler.go`, and change `DBaaSAuthzReconciler` and `DBaaSTenantReconciler` to have reference to the new type.

tchughesiv: sure, makes sense. let me move some things around.

tchughesiv: looking at it again i'm not sure i see a cleaner way than it's currently being done. while the `DBaaSTenantReconciler` only invokes `reconcileTenantAuthz`, that one function shares other `DBaaSAuthzReconciler` calls... which all rely on the `AuthorizationV1Client`.
https://github.com/RHEcosystemAppEng/dbaas-operator/blob/75e23665b402201d173fba79b21391cb2829860c/controllers/dbaasauthz_controller.go#L184-L186

xieshenzh (approved):

jianrongzhang89 (approved):
Excellent work! LGTM.

PR #96, opened 2022-01-13
Update the code to install SBO using OLM Dependency. Fix SBO Binding Permission issue 

redhatHameed:
## Description

Change the approach of SBO installation using OLM Dependency see the file dependencies.yaml within bundle folder.
updated the version of SBO to use : ">=1.0.0"

Fix the below-seen permission issue of the Service binding operator into the DBaaS install namespace by adding RoleBinding which refers to service-binding-controller-role 

Screen Shot 2022-01-17 at 3 50 15 PM

updated SBO labels as per the latest operator version guideline. 

Jira issue:  
https://issues.redhat.com/browse/DBAAS-255  

## Verification Steps

1. Check out the PR code and make an image and install the DBaaS operator.
2. Make sure the SBO operator get installed successfully 
3. Create DBaaS connection
4. Install sample application for Crunchy or MongoDB based on the connection provider type.
5. Connect the application with connection from topology view
6. Check if the application successfully connected with the selected database.

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

- [ ] Code Review completed


redhatHameed:
@jeremyary @xieshenzh @tchughesiv can you please take a look at this PR, once get time thanks

controllers/reconcilers/servicebinding/servicebinding_reconciler.go
```diff
@@ -149,35 +140,11 @@ func (r *Reconciler) waitFoServiceBindingOperator(cr *v1.DBaaSPlatform, ctx cont
...
-	if err := ctrl.SetControllerReference(cr, csv, r.scheme); err != nil {
-		return v1.ResultFailed, err
-	}
-	if err := r.client.Update(ctx, csv); err != nil {
-		return v1.ResultFailed, err
-	}
-	return v1.ResultInProgress, nil
-}
-
 func (r *Reconciler) getServiceBindingCSV(cr *v1.DBaaSPlatform) *v1alpha1.ClusterServiceVersion {
```
tchughesiv, line 143: we can no longer assume we're the only ones using a specific install of service binding in a cluster. given this, is this function still necessary at all? it appears to only be used to delete the csv. Similarly, the function it's used in, `func (r *Reconciler) Cleanup` should no longer be necessary.

tchughesiv (changes requested):

controllers/reconcilers/servicebinding/servicebinding_reconciler.go
redhatHameed: agreed and it's not used anymore removed the getServiceBindingCSV and cleanup logic from `func (r *Reconciler) Cleanup`. Keep the empty function signature as its implementing the [PlatformReconciler](https://github.com/RHEcosystemAppEng/dbaas-operator/blob/793e507b429eddf482d5efa3f18a522275fc3b2e/controllers/reconcilers/reconciler.go)

xieshenzh (dismissed):

redhatHameed:
> has this deployment been tested in an env where servicebinding is already installed. what about in an env where the wrong version is installed? i'm not sure we can/should change the version that a cluster's using in that case. some things to consider here.

The current operator behaves like, if there is an already installed version for example version v.0.1.0 the DBaaS operator will not be able to install the latest version but I see some errors in OLM like

"level":"error","ts":1642177972.2787035,"logger":"controllers.operatorcondition-generator","msg":"Unable to find ClusterServiceVersion","request":"openshift-dbaas-operator/service-binding-operator.v1.0.0","error":"ClusterServiceVersion.operators.coreos.com \"service-binding-operator.v1.0.0\" not 

but there is one issue we need to handle related with labels settings, we have different labels of permission to set according to the SBO version 
for example if the version v.0.1.0 the label is service.binding/controller: "true" in v.0.11 onwards the label is :     servicebinding.io/controller: "true" can we set both labels?

tchughesiv:
@redhatHameed if we move towards OLM dependencies declaration, this is no longer and issue we have to solve.

redhatHameed:
> if we're moving toward the official service binding operator, it feels like we should also switch to using the official dependency feature of olm. based on my limited understanding, a new dependencies.yaml file would be added to our bundle with something like the following:
> 
> ```yaml
> dependencies:
>   - type: olm.package
>     value:
>       packageName: service-binding-operator
>       version: ">=v0.10.0"
> ```
> 
> This would take care of everything for us ... installs, upgrades, etc. The `servicebinding_reconciler.go` code could then be removed. thoughts?
> 
> https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md#bundle-dependencies https://olm.operatorframework.io/docs/concepts/olm-architecture/dependency-resolution/

@jeremyary your thoughts on this moving SBO operator installation using olm dependency

redhatHameed:
> @redhatHameed if we move towards OLM dependencies declaration, this is no longer an issue we have to solve.

Sounds good to me, but I think we still need to define the labels inside [dbaasconnection_viewer_role](https://github.com/RHEcosystemAppEng/dbaas-operator/blob/1c3cf8d137421632eb87e5da6356cc5dd701f24a/config/rbac/dbaasconnection_viewer_role.yaml#L7) based on the SBO version installed. We can define the two labels so should not be an issue

jeremyary:
@redhatHameed @tchughesiv I'm ok w/it, let's go for it. Some day, we hope they're all [trusted enough to be olm deps](https://gitlab.cee.redhat.com/service/managed-tenants/-/blob/main/docs/tenants/addons_dependencies.md#trusted-catalogs-list) :) To back up Tommy here, we should thoroughly test this in a fresh cluster, also in an SBO older version update cluster, etc. to make sure we're ok in scenarios we may encounter.

redhatHameed:
@tchughesiv @jeremyary 

OLM dependency will not work in our case. Because it installed SBO into RHODA/DBaaS installed namespace and we will face the same permission issue. The SBO only works properly into `openshift-operators` namespace.

bundle/manifests/dbaas-operator-dbaasconnection-viewer-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   creationTimestamp: null
   labels:
     service.binding/controller: "true"
```
tchughesiv, line 6: can we remove this label? deprecated?

redhatHameed: we can remove that if we stick with version >=1.0.0

redhatHameed: I think we need the version >=1.0.0 because there is another issue we have seen for using deploymentconfig kind, which is not supported in version 0.10.0 
cc: @jeremyary

tchughesiv (changes requested):
lgtm, just rebase and we can merge

redhatHameed:
> lgtm, just rebase and we can merge

Done.

controllers/reconcilers/constants.go
```diff
@@ -16,5 +16,5 @@ const (
 	CONSOLE_TELEMETRY_PLUGIN_SEGMENT_KEY     = "qejcCDG37ICCLIDsM1FcJDkd68hglCoK"
 	MONGODB_ATLAS_CSV                        = "mongodb-atlas-kubernetes.v0.1.0"
 	CRUNCHY_BRIDGE_CSV                       = "crunchy-bridge-operator.v0.0.2-dev"
-	SERVICE_BINDING_CSV                      = "service-binding-operator.v0.10.0"
+
```
tchughesiv, line 19: appears you may need to run a `make fmt` and commit changes... one of the checks is failing.

tchughesiv (approved):
lgtm

PR #105, opened 2022-01-28
added subscription environment variable for SYNC-PERIOD-MIN

redhatHameed:
The operator does not allow users to change environment variables. All changes get reverted. the possible solution is to add into subscription config: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md#subscription-config

add filed SyncPeriod into DBaaS-platform CR to make it configurable. 

## Type of change

- [ ] Code Review completed


api/v1alpha1/dbaasplatform_types.go
```diff
@@ -41,7 +41,7 @@ const (
 
 // DBaaSPlatformSpec defines the desired state of DBaaSPlatform
 type DBaaSPlatformSpec struct {
-	Name string `json:"name,omitempty"`
+	SyncPeriod string `json:"syncPeriod,omitempty"`
```
tchughesiv, line 44: would it be a good idea to make this an integer instead? then convert to string in code?

redhatHameed: make sense, added validation as well. updated

config/crd/bases/dbaas.redhat.com_dbaasplatforms.yaml
```diff
@@ -36,8 +36,10 @@ spec:
           spec:
             description: DBaaSPlatformSpec defines the desired state of DBaaSPlatform
             properties:
-              name:
-                type: string
+              syncPeriod:
+                maximum: 180
```
tchughesiv, line 40: any reason 180 is the max here? just curious

tchughesiv (changes requested):
can we add a couple unit tests that verify no ENV is set when `cr.Spec.SyncPeriod` is empty? and another that verifies one IS set? also, what happens if an empty `SYNC-PERIOD-MIN` var is set? does that mean no refresh?

api/v1alpha1/dbaasplatform_types.go
```diff
@@ -41,7 +41,9 @@ const (
 
 // DBaaSPlatformSpec defines the desired state of DBaaSPlatform
 type DBaaSPlatformSpec struct {
-	Name string `json:"name,omitempty"`
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=180
+	SyncPeriod int `json:"syncPeriod,omitempty"`
```
tchughesiv, line 46: can we add a description here?

tchughesiv (changes requested):

config/crd/bases/dbaas.redhat.com_dbaasplatforms.yaml
jianrongzhang89: @redhatHameed you can set default: 180 instead of maximum.

redhatHameed: @jianrongzhang89 Thanks default already set by provider operator.

@tchughesiv added 24h max value

api/v1alpha1/dbaasplatform_types.go
redhatHameed: added

redhatHameed:
> can we add a couple unit tests that verify no ENV is set when `cr.Spec.SyncPeriod` is empty? and another that verifies one IS set? also, what happens if an empty `SYNC-PERIOD-MIN` var is set? does that mean no refresh?

added unit test.

added if check if empty/nil cr.Spec.SyncPeriod don't set environment variable

tchughesiv (approved):
Lgtm

controllers/dbaasplatform_controller_test.go
```diff
@@ -0,0 +1,48 @@
...
+				Labels:    map[string]string{"managed-by": "dbaas-operator"},
+			},
+
+			Spec: dbaasv1alpha1.DBaaSPlatformSpec{
+
+				SyncPeriod: nil,
+			},
+		}
+		BeforeEach(assertResourceCreation(cr))
+		AfterEach(assertResourceDeletion(cr))
```
tchughesiv, line 45: Sorry, Can we check the resulting subscription here? See if env var is set or not

tchughesiv (changes requested):

controllers/dbaasplatform_controller_test.go
redhatHameed: This required a complete platform installations unit test as currently not added. I have given try but there are some other issues faced, also we have dependencies on platform installation on openshift real environment and installation never completed see logs:
https://gist.github.com/redhatHameed/2ca684b6488b77193ab77b0591c933fb

tchughesiv (approved):
lgtm

PR #135, opened 2022-03-09
WIP: Changes related to metrics exporter

rchikatw:
## Description
Expected: Needs to run metrics services as in new pod. 

Added initial changes to expose metrics to prometheus. 

Notes wrt. folder structure.
1. Added metrics folder keep changes related to exporter separately. 
2. Will be setting all the metrics value in metrics/collector folder. 
3.  All the configuration files added in metrics/deploy folder. 

Please suggest any change required to make improve the code as well as design. Would love to address your input.


metrics/internal/deploy/rbac.yaml
```diff
@@ -0,0 +1,87 @@
...
+metadata:
+  name: dbaas-metrics-exporter
+rules:
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - '*'
+  verbs:
+  - '*'
+- apiGroups:
```
rchikatw, line 13: ignore this group i have to remove this

metrics/internal/deploy/rbac.yaml
```diff
@@ -0,0 +1,87 @@
...
+  - ceph.rook.io
```
rchikatw, line 14: have to remove this group as its not needed

redhatHameed:
@rchikatw can you add verification steps, how to verify/test this PR. Thanks

jeremyary:
> @rchikatw can you add verification steps, how to verify/test this PR. Thanks

this is more or less at my request so that we can sync on on-going issues. We should, though, convert this to a Draft.

jeremyary:
Ok, we should be in a non-notification Draft state now. Rohan and I will use this to work out some RBAC issues, but for others that find their way here, no need to review at this time (unless you want to wade in on the RBAC setup ;) )

redhatHameed:
@rchikatw can you close this PR as the we are going approach as operator-sdk recommendation for metrics

rchikatw:
Closing this PR as the another approach is considered.

PR #152, opened 2022-03-25
add label patch for other providers

tchughesiv:
related to https://github.com/jeremyary/ccapi-k8s-operator/pull/1

### DIFF
--- a/controllers/dbaas_reconciler.go
+++ b/controllers/dbaas_reconciler.go
@@ -34,8 +34,9 @@ import (
 var (
 	InstallNamespaceEnvVar = "INSTALL_NAMESPACE"
 	inventoryNamespaceKey  = ".spec.inventoryNamespace"
-	typeLabelValue         = "credentials"
-	typeLabelKeyMongo      = "atlas.mongodb.com/type"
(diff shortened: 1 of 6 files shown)


controllers/dbaas_reconciler.go
```diff
@@ -35,6 +35,7 @@ var (
 	InstallNamespaceEnvVar = "INSTALL_NAMESPACE"
 	inventoryNamespaceKey  = ".spec.inventoryNamespace"
 	typeLabelValue         = "credentials"
+	typeLabelKey           = "type"
```
jianrongzhang89, line 38: Would it be better to have a dbaas specific label similar to o mongo, such as "dbaas.redhat.com/type" to avoid conflicts?

tchughesiv: what kind of conflict?

jianrongzhang89: "type" seems to be a generic label that people may use for other purposes. Use a dbaas type label will avoid such "conflict"

tchughesiv: done

PR #159, opened 2022-04-11
[DBAAS-165] leverage RELATED_IMAGE functionality for image references

tchughesiv:
- removes hard-coded, release-specific configs... moves images and versions to env vars. should be set in `config/default/manager-env-images.yaml`
 - should env var(s) be missing, it will use default(s) via embedded yaml file
 - option to use image digests in csv, w/ `make bundle-w-digests`. example of result - https://gist.github.com/tchughesiv/6b0d19518797bfe6d138b0471e45100a
 - upgrade to operator-sdk v1.20.1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,6 +166,8 @@ catalog-update:
 	-oc delete catalogsource mongodb-atlas-catalogsource -n openshift-marketplace
 	-oc delete catalogsource crunchy-bridge-catalogsource -n openshift-marketplace
 	-oc delete catalogsource ccapi-k8s-catalogsource -n openshift-marketplace
+	-oc delete catalogsource observability-catalogsource -n openshift-marketplace
+	-oc delete catalogsource rds-provider-catalogsource -n openshift-marketplace
 	 oc apply -f config/samples/catalog-source.yaml
 
 deploy-sample-app:
@@ -209,11 +211,19 @@ rm -rf $$TMP_DIR ;\
 }
 endef
 
-.PHONY: bundle
-bundle: manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
+.PHONY: sdk-manifests
+sdk-manifests: manifests kustomize ## Generate bundle manifests and metadata.
 	operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
+
+.PHONY: bundle
+bundle: sdk-manifests ## Generate bundle manifests, then validate generated files.
 	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --manifests --version $(VERSION) $(BUNDLE_METADATA_OPTS)
+	operator-sdk bundle validate ./bundle
+
+.PHONY: bundle-w-digests
+bundle-w-digests: sdk-manifests ## Generate bundle manifests w/ image digests, then validate generated files.
+	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --manifests --use-image-digests --version $(VERSION) $(BUNDLE_METADATA_OPTS)
 	operator-sdk bundle validate ./bundle
 
 .PHONY: bundle-build
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -10,7 +10,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=dbaas-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.10.0+git
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.19.0+git
 LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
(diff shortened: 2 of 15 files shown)


tchughesiv:
@jeremyary i was originally intended to work this for 0.2.0 ... do we still want to attempt or push to next release?

jeremyary:
We'll collectively touch base and figure out our progress today, so pending that if we end up with extra cycles, but next release is fine.

controllers/reconcilers/constants.go
```diff
@@ -78,7 +80,7 @@ var InstallationPlatforms = map[dbaasv1alpha1.PlatformsName]dbaasv1alpha1.Platfo
 		Name:           CRUNCHY_BRIDGE_NAME,
 		CSV:            CRUNCHY_BRIDGE_CSV,
 		DeploymentName: CRUNCHY_BRIDGE_DEPLOYMENT,
-		Image:          CRUNCHY_BRIDGE_CATALOG_IMG,
+		Image:          os.Getenv(CRUNCHY_BRIDGE_CATALOG_ENV_VAR),
```
redhatHameed, line 83: what do you think if we can get the image from constant.go if there is no image into environment variable.

tchughesiv: yea, been thinking through options w/ this... idk. issue w/ that is we'd have to changes images in multiple locations which isn't ideal.

tchughesiv: i might explore embedding a reference file or something that we can also use w/ kustomize during bundle generation

jianrongzhang89: @tchughesiv Moving the images to the env vars gives us flexibility. However other variables are still on constants.go file so the related info is now in two places.  What about moving all of them into a config map?

tchughesiv: configmap is not the supported method through olm / osbs and would lead to issues w/ upgrades or installs if an existing configmap with the wrong images were in place

tchughesiv: basically a support/maintenance nightmare and messed up installations

jianrongzhang89: @tchughesiv OK. Thanks

jianrongzhang89 (approved):

controllers/reconcilers/constants.go
```diff
@@ -1,7 +1,12 @@
 package reconcilers
```
redhatHameed, line 1: change file name constants.go-> settings.go WDYT ?

Makefile
```diff
@@ -209,13 +211,21 @@ rm -rf $$TMP_DIR ;\
 }
 endef
 
-.PHONY: bundle
-bundle: manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
+.PHONY: sdk-manifests
```
redhatHameed, line 214: why this change ?

bundle.Dockerfile
```diff
@@ -10,7 +10,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=dbaas-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.10.0+git
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.19.0+git
```
redhatHameed, line 13: operator sdk mentioned in readme operator-sdk v1.20.1 in docker file here it showing `v1.19.0+git` do we need to change in readme ?

config/default/manager-env-images.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: apps/v1
+kind: Deployment
```
redhatHameed, line 2: where we need to change the image version in future, I mean which also can we add some description in readme for changing the setting the version

bundle.Dockerfile
tchughesiv: it's common w/ the sdk that this gets set at a version below the binary's release. this is actually the version that is set in the csv by sdk 1.20.1.

Makefile
tchughesiv: so that both `make bundle` and `make bundle-w-digests` can leverage the same commands

config/default/manager-env-images.yaml
tchughesiv: going forward, all third-party images and versions will be set in this file... including our dynamic plugin and observability images.

tchughesiv: ill add a note at the top of this file for sure.

redhatHameed: Got it thanks

tchughesiv: done

redhatHameed (approved):
Verified the params are passing to environment variables and operator is installing accordingly.
Screen Shot 2022-08-04 at 1 26 36 PM

@tchughesiv great works.  can we change the environment variables values from console ?  as  my understanding is that operator will change back these values when it get reconciled, the reason to ask just incase if we want to try new provider version without changing the code.

tchughesiv:
@redhatHameed once deployed by olm (via bundle) those settings are immutable. that's a good thing in production. that said, since they're loaded via the embedded file... those could be changed and used if run locally.

tchughesiv:
@redhatHameed the biggest benefit here is should any provider changes be required... versions, images, etc... we can just build a new bundle. we don't have to build/release a new operator image.

jeremyary (approved):

PR #175, opened 2022-05-05
[DBAAS-417] tenancy refactor

tchughesiv:

## Description
move away from cluster-scoped DBaaSTenant object -
 - switch to new namespaced `DBaaSPolicy` CR. It enables admin capabilities within a namespace and sets default inventory policy. Policy defaults can be overridden on a per-inventory basis.
 - leverage `ResourceQuota` to limit to 1 `DBaaSPolicy` per  namespace
 - add ability to disable instance/db provisioning at the inventory or policy level through the `spec.disableProvisions` field

https://issues.redhat.com/browse/DBAAS-417

## Type of change

- [ ] Code Review completed


api/v1alpha1/dbaasprovider.go
```diff
@@ -29,23 +29,29 @@ const (
...
 	// DBaaS condition messages
 	MsgProviderCRStatusSyncDone      string = "Provider Custom Resource status sync completed"
 	MsgProviderCRReconcileInProgress string = "DBaaS Provider Custom Resource reconciliation in progress"
 	MsgInventoryNotReady             string = "Inventory discovery not done"
-	MsgTenantNotFound                string = "Failed to find DBaaS tenants"
+	MsgInventoryNotProvisionable     string = "Inventory provisioning not allowed"
+	MsgConfigNotFound                string = "Failed to find an active Config"
+	MsgConfigReady                   string = "Config is active"
 	MsgInvalidNamespace              string = "Invalid connection namespace for the referenced inventory"
+	MsgConfigNotReady                string = "Another active Config already exists"
```
jeremyary, line 54: It's possible we could end up showing these via UX similarly to how we've done with the underlying cause of Provider Account fetch fails. Could we use the full "DBaaSConfig" name here that would match the CR if they wanted to dig in and perhaps just sanity-check the grammar/sentiment on these to make sure they're understandable to the end user? WDYT?

bundle/manifests/dbaas.redhat.com_dbaasconfigs.yaml
```diff
@@ -0,0 +1,140 @@
...
+            type: object
+          status:
+            description: DBaaSConfigStatus defines the observed state of DBaaSConfig
+            properties:
+              conditions:
+                items:
+                  description: "Condition contains details for one aspect of the current
+                    state of this API Resource. --- This struct is intended for direct
+                    use as an array at the field path .status.conditions.  For example,
+                    type FooStatus struct{     // Represents the observations of a
```
jeremyary, line 65: Foo conditions always messing up my FooStatus XD

config/samples/catalog-source.yaml
```diff
@@ -5,5 +5,5 @@ metadata:
   namespace: openshift-marketplace
 spec:
   sourceType: grpc
-  image: quay.io/ecosystem-appeng/dbaas-operator-catalog:v0.1
-  displayName: DBaaS Operator
\ No newline at end of file
+  image: quay.io/tchughesiv/dbaas-operator-catalog:v0.2.0
```
jeremyary, line 8: I'm guessing intentional change to personal reference since I'm not pushing catalog on release. Would you prefer I included that step instead? Or you don't mind this?

controllers/dbaas_reconciler.go
```diff
@@ -24,50 +24,15 @@ import (
...
-		return false
-	},
-	DeleteFunc: func(e event.DeleteEvent) bool {
-		return false
-	},
-	GenericFunc: func(e event.GenericEvent) bool {
-		return false
-	},
-}
-
```
jeremyary, line 70: just general housekeeping in knocking out unused func here and below?

controllers/reconcilers/constants.go
```diff
@@ -40,7 +39,7 @@ const (
 	COCKROACHDB_CHANNEL     = "alpha"
 
 	// DBAAS_DYNAMIC_PLUGIN
-	DBAAS_DYNAMIC_PLUGIN_IMG          = "quay.io/ecosystem-appeng/dbaas-dynamic-plugin:0.2.0"
+	DBAAS_DYNAMIC_PLUGIN_IMG          = "quay.io/tchughesiv/dbaas-dynamic-plugin:0.2.0"
```
jeremyary, line 42: this one feels accidental, should reference `ecosystem-appeng` for installations

bundle/manifests/dbaas.redhat.com_dbaasconfigs.yaml
tchughesiv: Ha, believe it or not this comes from an imported k8s module for the Condition type -
https://github.com/kubernetes/apimachinery/blob/v0.22.1/pkg/apis/meta/v1/types.go#L1384-L1387

This descrip exists in most of our dbaas status crds

config/samples/catalog-source.yaml
tchughesiv: leftover dev env changes... will fix

controllers/reconcilers/constants.go
tchughesiv: leftover dev env changes... will fix

controllers/dbaas_reconciler.go
tchughesiv: correct, these are used by the `dbaastenant` and `dbaastenantauthz` controllers... both of which go away with this change.

jeremyary (dismissed):

api/v1alpha1/dbaasprovider.go
tchughesiv: added existing active config name to message

jeremyary (dismissed):

PROJECT
```diff
@@ -41,10 +41,11 @@ resources:
   version: v1alpha1
 - api:
     crdVersion: v1
+    namespaced: true
   controller: true
   domain: redhat.com
   group: dbaas
-  kind: DBaaSTenant
+  kind: DBaaSConfig
```
jianrongzhang89, line 48: "DBaaSConfig" seems a rather technical term. What about changing it to something like "ProviderAccountPolicy" that is more meaningful for end-users?

bundle/manifests/dbaas-operator.clusterserviceversion.yaml
```diff
@@ -288,6 +266,22 @@ spec:
           - list
           - update
           - watch
+        - apiGroups:
+          - ""
+          resources:
+          - resourcequotas
```
jianrongzhang89, line 272: Using resource quota to guarantee there is only one dbaasconfig seems an overkill. It can also lead to confusion. What if the admin user updates the resource quota accidentally? Will the dbaas operator continue to work?

Can such validation be replaced with a webhook instead?

PROJECT
tchughesiv: @jianrongzhang89 thanks for the review!

I'm not attached to that object kind and definitely open to changing it. That said, moving away from the "DBaaSxxx" prefix/format would likely need broader consensus from the team considering every other operator object uses this format.

Unfortunately we can't keep `DBaaSTenant` as the kind, since it already exists as a cluster-scoped object. I attempted this transition in a lab and it failed consistently.

IMO, the kind name should try to reflect the idea that it’s a necessary object in order to use rhoda within that namespace. More broadly, it should reflect that this object can be used to configure rhoda for said namespace. Longer term, additional options may be added to this object. 

A couple other ideas I considered -
 - DBaaSMgmt
 - DBaaSInit

bundle/manifests/dbaas-operator.clusterserviceversion.yaml
tchughesiv: > Can such validation be replaced with a webhook instead?

I thought so and this was actually the first thing I tried… it just wasn’t dependable enough. I was consistently able to sneak additional configs through the webhook. For example, using a List to create multiples at one time. Additionally it was often sluggish in response.

This caused me to research other options and that’s when I discovered openshift already supports this type of object limit validation natively via the use of `ResourceQuotas`. It’s worth noting I’ve yet to sneak through another config when using RQs and there’s no apparent effect to cluster response.


> What if the admin user updates the resource quota accidentally?

A `ResourceQuota` is created only when a user creates a `DBaaSConfig`. The operator sets this `DBaaSConfig` as the owner of it’s `ResourceQuota`. Any changes made to the owned RQ will trigger a reconcile by the operator, which will “fix” the RQ. I’ve tested this many times and it’s immediate. If the RQ is deleted, it will be recreated.


> Will the dbaas operator continue to work?

Yes, and the operator doesn’t actually require RQs to function properly. In the rare event that a user were to remove an RQ’s ownerrefs and make changes -
 - Should they change the hard limit to 0, the existing `DBaaSConfig` would continue to exist. Nothing would change. RQs don’t remove objects if a limit is exceeded.
 - Should they delete the limit setting, the RQ itself, or change the hard limit to something greater than 1, the user would then be able to create multiple configs in that namespace. In this scenario, the operator would assign only one config per namespace as the “active” config, in its status. This config is the only one the operator will reference during inventory/connection/instance reconciliations. If that config is removed, the operator will then assign another as active.


> Using resource quota to guarantee there is only one dbaasconfig seems an overkill. It can also lead to confusion.

I’m not sure I agree. `ResourceQuotas` are native to openshift, same as deployments or secrets, and are designed for this exact kind of use-case. 

It was/is my opinion that allowing multiple configs to exist within a namespace would be more confusing to users than creating a limit with a simple RQ. Without this limit, multiple configs could exist per namespace and a user would have to check each config to determine which of them was active before making changes… or to troubleshoot. It would also cause the user to believe that, with the creation of each config, the operator should take some kind of action… which is not true for this object.

Using RQs prevents these scenarios and sends a message to the user (without them having to read the spec) that only one config is needed per namespace. Why allow additional configs within a namespace if we only use one?

tchughesiv: All of that said, if the team thinks RQs are overkill… we could just allow users to create as many configs as they like and rely entirely on the operator’s “one active” logic. Again, this seems more confusing to me and the technology exists natively to avoid it.

PROJECT
jianrongzhang89: What about DBaaSPPolicy?

tchughesiv: interesting... i def like it. @jeremyary thoughts? should we maybe put it to a vote w/ the broader team?

tchughesiv: agree... i'll roll w/ policy

api/v1alpha1/dbaaspolicy_types.go
```diff
@@ -36,25 +36,25 @@ type DBaaSConfigStatus struct {
 //+kubebuilder:printcolumn:name="Active",type=string,JSONPath=`.status.conditions[0].status`
 //+kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 
-//+operator-sdk:csv:customresourcedefinitions:displayName="DBaaSConfig"
-// DBaaSConfig is the Schema for the dbaasconfigs API
-type DBaaSConfig struct {
+//+operator-sdk:csv:customresourcedefinitions:displayName="DBaaSPolicy"
```
tchughesiv, line 39: @jeremyary @jianrongzhang89 do we want to set "Provider Account Policy" here... to stay in-line w/ https://github.com/RHEcosystemAppEng/dbaas-operator/blob/856ac28ad79fc0fbeef90b1b40dfa969d226584d/api/v1alpha1/dbaasinventory_types.go#L40

tchughesiv: i think this makes sense to do... will move forward with the change. let me know if you disagree

bundle/manifests/dbaas-operator.clusterserviceversion.yaml
jianrongzhang89: Thanks @tchughesiv for clarification. This is fine to me.

api/v1alpha1/dbaaspolicy_types.go
jianrongzhang89: @tchughesiv sounds good

config/manifests/bases/dbaas-operator.clusterserviceversion.yaml
```diff
@@ -18,6 +18,11 @@ spec:
   apiservicedefinitions: {}
   customresourcedefinitions:
     owned:
+    - description: DBaaSConfig is the Schema for the dbaasconfigs API
```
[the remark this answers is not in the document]
tchughesiv: No, just one. Still working on changes.

api/v1alpha1/dbaasprovider.go
```diff
@@ -132,6 +141,17 @@ type LocalObjectReference struct {
 	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
 }
 
+// DBaaSInventoryConfigs configures inventories
+type DBaaSInventoryConfigs struct {
```
jianrongzhang89, line 145: The name DBaaSInventoryConfigs is plural and makes one think this is used for an array. What about changing it to DBaaSInventoryPolicy?

tchughesiv: this makes sense to me... changing

tchughesiv: i've made the change and added a better description of the policy object.
```
    - description: DBaaSPolicy enables admin capabilities and sets the default policy
        for inventories within a namespace. Default policy can be overridden on a
        per-inventory basis.
```

thanks!

jianrongzhang89: Thanks!

jianrongzhang89 (approved):
LGTM

jeremyary (approved):

PR #186, opened 2022-05-27
[DBAAS-608] a referenced secret must reside in the inventory's namespace

tchughesiv:
https://issues.redhat.com/browse/DBAAS-608


api/v1alpha1/dbaasprovider.go
```diff
@@ -146,6 +146,11 @@ type Instance struct {
 	InstanceInfo map[string]string `json:"instanceInfo,omitempty"`
 }
 
+type Name struct {
```
jianrongzhang89, line 149: Name => ResourceRef?

tchughesiv: switched instead to the existing `corev1.LocalObjectReference`

api/v1alpha1/dbaasprovider.go
```diff
@@ -120,9 +120,16 @@ type CredentialField struct {
...
+	// DBaaSProvider CR (CredentialFields key). The Secret must exist within the same namespace
+	// as the Inventory.
+	CredentialsRef *LocalObjectReference `json:"credentialsRef"`
+}
+
+// LocalObjectReference contains enough information to let you locate the
+// referenced object inside the same namespace.
+type LocalObjectReference struct {
+	// Name of the referent.
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
```
jeremyary, line 132: any version upgrade implications come to mind if you've thought on that yet?

tchughesiv: for inventories created w/ UI, we already default to secrets in same namespace. So, nothing changes with an upgrade. The `namespace` field disappears and the operator continues to look in the same place as before. The only impact would be to inventories referencing secrets in a different namespace... which is the exact situation we're aiming to avoid anyway.

jeremyary (approved):

PR #192, opened 2022-06-15
Added mandatory labels for DockerFile

fOO223Fr:
## Description
These labels in DockerFile are needed as mentioned [here](https://source.redhat.com/groups/public/container-build-system/container_build_system_wiki/guide_to_layered_image_build_service_osbs#jive_content_id_The_Dockerfile)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Labels
-LABEL com.redhat.component="dbaas-operator-container"
-LABEL name="dbaas-operator"
-LABEL version="1.0"
+LABEL com.redhat.component="rhoda-operator-container"
+LABEL name="rhoda-operator"
+LABEL version="0.2"
 
 # Build the manager binary
 FROM registry.access.redhat.com/ubi8:8.5 AS builder
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,8 @@
+# Labels
+LABEL com.redhat.component="dbaas-operator-container"
+LABEL name="dbaas-operator"
+LABEL version="1.0"
+
 # Build the manager binary
 FROM registry.access.redhat.com/ubi8:8.5 AS builder
 


Dockerfile
```diff
@@ -1,3 +1,8 @@
+# Labels
+LABEL com.redhat.component="dbaas-operator-container"
+LABEL name="dbaas-operator"
+LABEL version="1.0"
```
tchughesiv, line 4: should this version match the release version being built?
if so, the build scripting should be incorporated to dynamically set this at build time... maybe through use of container build ARGs?

tchughesiv (changes requested):

Dockerfile
fOO223Fr: @tchughesiv yes i think this is the release version so for our case it would be "0.5" which is "major.minor" . This version will also go to the branch names for example "rhoda-1.0-rhel-8" ?
I agree with build scripting but for the moment we are trying to do manually/absolute values and once we have successful release in this way then we know that productization is working. Having that confirmation I would look for automation around it. I do have some other automation planned too

tchughesiv: in our case, i think we'd want `0.2.0` to match our main version set at -
https://github.com/RHEcosystemAppEng/dbaas-operator/blob/856ac28ad79fc0fbeef90b1b40dfa969d226584d/Makefile#L6

wrt dist-git branches... i think we need to discuss as a group and make a decision

tchughesiv (changes requested):
@fOO223Fr gh actions throwing an error during build -
`no FROM statement found`
https://github.com/RHEcosystemAppEng/dbaas-operator/runs/6993714792?check_suite_focus=true#step:3:54

does this build successfully locally?

Dockerfile
```diff
@@ -1,3 +1,8 @@
+# Labels
+LABEL com.redhat.component="rhoda-operator-container"
+LABEL name="rhoda-operator"
+LABEL version="0.2"
+
```
tchughesiv, line 5: i believe these should be moved down to be between these items, under the secondary FROM image -
https://github.com/RHEcosystemAppEng/dbaas-operator/pull/192/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R35-R37

this should fix your build issues and will also ensure the final image contains these labels. going forward, it's always a good idea to test your changes locally before submitting PR.

tchughesiv (changes requested):

PR #196, opened 2022-06-30
 creating the observability operator MonitoringStack CR 

redhatHameed:
Signed-off-by: Abdul Hameed <ahameed@redhat.com>

## Type of change

- [ ] Code Review completed


redhatHameed:
@[rewantsoni](https://github.com/rewantsoni) opened the draft PR for creating the CR for observability operator from RHODA.

controllers/reconcilers/providers_installation/providers_installation_reconciler.go
```diff
@@ -60,6 +60,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, cr *v1.DBaaSPlatform, status
 	if status != v1.ResultSuccess {
 		return status, err
 	}
+
+	if r.config.Name == reconcilers.OBSERVABILITY_NAME {
```
rewantsoni, line 64: When we no longer need to install the Observability operator, we would have to move this somewhere else, can we move it so that we don't have to make changes to our code when we have the Observability Operator available already in the cluster?

controllers/reconcilers/providers_installation/observability-operator-cr.yaml
```diff
@@ -0,0 +1,29 @@
...
+  name: dbaas-operator-mso
+spec:
+  logLevel: debug
+  resourceSelector:
+    matchLabels:
+      app: dbaas-prometheus
+  prometheusConfig:
+    remoteWrite:
+      - bearerToken: >-
+          <bearer_token>
```
rewantsoni, line 13: How are we going to store and retrieve this token, observatorium_host and tenant_id?

controllers/reconcilers/providers_installation/providers_installation_reconciler.go
rewantsoni: I think we can test it if we create the cluster, deploy the observability operator and then install the DBaaS operator. If it doesn't create the operator again and doesn't throw any error, we should be good to go.

Yes, I meant in a different package, according to me we can move it into platform_reconciler, that is our main package where we do all our initial setup. We could create a new reconciler but that doesn't make sense if we are just deploying the Observability CR. We can handle the cleanup in platform_reconciler as well.

redhatHameed: > I think we can test it if we create the cluster, deploy the observability operator and then install the DBaaS operator. If it doesn't create the operator again and doesn't throw any error, we should be good to go.
 

> Yes, I meant in a different package, according to me we can move it into platform_reconciler, that is our main package where we do all our initial setup. We could create a new reconciler but that doesn't make sense if we are just deploying the Observability CR. We can handle the cleanup in platform_reconciler as well.

The reason to call the MonitoringStack CR creation here, to make sure the operator installation(subscription, catalog source, deployment ready ) before creating the CR, as all resource creation take place within that package/file for operators, and for CR creating the logic into different file for easy removal in the future [file](https://github.com/RHEcosystemAppEng/dbaas-operator/blob/742627ae86a3ca85e9cc95c73633bdd1f61e46ae/controllers/reconcilers/providers_installation/observability-operator.go), 
for cleanup the CR already set owner reference of DBaaSPlatform so cleanup/deletion of the CR will be performed automatically when  DBaaSPlatform cr deleted.

PR #197, opened 2022-07-05
move to go 1.17

tchughesiv:


jianrongzhang89:
Since go1.18 is available, is there a reason why we do not upgrade to 1.18 directly?

tchughesiv:
@jianrongzhang89 there's not a 1.18 registry.access.redhat.com/ubi8/go-toolset image yet. my intention is to get back to leveraging RH bits for builds so that we can do proper release builds through osbs in the future. that's generally a requirement to use osbs.

jianrongzhang89:
> 

That's good. Thanks

Makefile
```diff
@@ -130,7 +130,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
 docker-build: test ## Build docker image with the manager.
-	$(CONTAINER_ENGINE) build -t ${IMG} .
+	$(CONTAINER_ENGINE) build --pull -t ${IMG} .
```
jianrongzhang89, line 133: In line #207, "go get" has been deprecated and should be replaced with "go install".

tchughesiv: i've tried moving to `go install` but get the following error - 
```
go install: sigs.k8s.io/kustomize/kustomize/v3@v3.8.7 (in sigs.k8s.io/kustomize/kustomize/v3@v3.8.7):
	The go.mod file for the module providing named packages contains one or
	more exclude directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```
i think we will have to wait a while longer before we can change that. luckily, get still works fine.

jeremyary (approved):

PR #212, opened 2022-07-25
added environment variables for dev observatorium, added creation timestamp label into metrics and update version 0.0.13 for observability operator

redhatHameed:
## Description
1- Added params as a environment variable for dev observatorium to read value from secret and pass to observability cr. 
2- Upgraded the version of Observability operator to 0.0.13
3-Added label for creation timestamp into inventory, connection, and instance status metrics, to know when particular object was created.

## Verification Steps

1- build the operator with this PR and install it
2- configure remote write observtorium by creating the secret with valid value see example of secret [here](https://github.com/redhatHameed/dbaas-operator/blob/obs-env/config/samples/secret-dbaas-operator-dev-prom-remote-write.yaml)
3- verify that all metrics are pushing to locall prometheus and remote observatorium
4- Verify the observability version installed with version 0.0.13
5- verify the metrics contain label creation-timestamp 

## Type of change

- [ ] Code Review completed


jianrongzhang89:
@redhatHameed What about using separate metrics for the creation time stamp itself instead of adding it to each metric (like what kube state metrics does)?  This is because there are multiple metrics for each inventory/instance/connection CR, and in the future, we may want to add additional metrics.

redhatHameed:
> @redhatHameed What about using separate metrics for the creation time stamp itself instead of adding it to each metric (like what kube state metrics does)? This is because there are multiple metrics for each inventory/instance/connection CR, and in the future, we may want to add additional metrics.

At the moment, we have requirements like, when inventory/connection/instance created, what is the status, who is the provider, so with the single metrics we are getting all the required information. 
if we need to define extra metrics for creation time, then we have to query multiple metrics to get information.

redhatHameed:
> @redhatHameed Just realized you updated the PR and removed the creation timestamp related changes. Could you update the PR description?

it's there not removed just updated to add timestamp to request duration metrics :)

jianrongzhang89:
> 

OK

bundle/manifests/dbaas-operator.clusterserviceversion.yaml
```diff
@@ -442,6 +442,36 @@ spec:
...
+                    secretKeyRef:
+                      key: RHOBS_TOKEN
+                      name: dbaas-operator-dev-prom-remote-write
+                      optional: true
+                - name: ADDON_NAME
+                  valueFrom:
+                    secretKeyRef:
+                      key: ADDON_NAME
+                      name: dbaas-operator-dev-prom-remote-write
+                      optional: true
```
tchughesiv, line 474: how is this secret created? would these additions make it a requirement for the operator to start?
what happens if it doesn't exist?

tchughesiv: nm, i see now they're marked as `optional`

controllers/reconcilers/providers_installation/observability-operator.go
```diff
@@ -71,13 +64,155 @@ func (r *Reconciler) createObservabilityCR(cr *dbaasv1alpha1.DBaaSPlatform, ctx
 		}
 	} else if len(monitoringStackList.Items) == 1 {
 		monitoringStackCR = &monitoringStackList.Items[0]
+		if config.RemoteWritesURL != "" && config.AuthType != "" && config.AddonName != "" {
+			prometheusConfig, _ := r.setPrometheusConfig(ctx, config, monitoringStackCR.Namespace)
+			monitoringStackCR.Spec.PrometheusConfig = prometheusConfig
+			if _, err := controllerutil.CreateOrUpdate(ctx, r.client, monitoringStackCR, func() error {
```
jianrongzhang89, line 70: If there is already one monitoring stack CR, why do you need to create a new one? If so it would end with two CRs?

config/samples/secret-dbaas-operator-dev-prom-remote-write.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: dbaas-operator-dev-prom-remote-write
+stringData:
+  AUTH_TYPE: dex
+  RHOBS_API_URL: https://observatorium-observatorium.apps.cluster.openshift.com/api/metrics/v1/test2/api/v1/receive
+  ADDON_NAME: dbaas-operator-dev
+  RH_SSO_TOKEN_ENDPOINT: https://sso.com
+  RHOBS_TOKEN: <token>
+  prom-remote-write-config-id: test
+  prom-remote-write-config-secret: test
+  rhobs-audience: test
```
jianrongzhang89, line 13: Could you add a new line?

controllers/reconcilers/providers_installation/observability-operator.go
redhatHameed: I am trying to update the same CR- `monitoringStackCR = &monitoringStackList.Items[0]` the reason to update the same CR is  if any value change in environment variable, I want update the value accordingly

config/samples/secret-dbaas-operator-dev-prom-remote-write.yaml
redhatHameed: done

api/v1alpha1/dbaasplatform_types.go
```diff
@@ -49,15 +49,25 @@ const (
...
+	Name                string
+	CSV                 string
+	DeploymentName      string
+	Image               string
+	PackageName         string
+	Channel             string
+	DisplayName         string
+	Envs                []v1.EnvVar
+	Type                PlatformsType
+	ObservatoriumConfig ObservatoriumConfig
```
xieshenzh, line 61: `PlatformConfig` is supposed to be generic, so can we make this field not so specific to the observability operator?
For example, change the name to `OtherConfigs` and the type to be `map[string]string` or some new interface.

xieshenzh (changes requested):

api/v1alpha1/dbaasplatform_types.go
redhatHameed: make sense, updated `OtherConfigs   map[ConfigName]OtherConfig` have a review again thanks

api/v1alpha1/dbaasplatform_types.go
```diff
@@ -58,6 +61,16 @@ type PlatformConfig struct {
 	DisplayName    string
 	Envs           []v1.EnvVar
 	Type           PlatformsType
+	OtherConfigs   map[ConfigName]OtherConfig
```
xieshenzh, line 64: I don't think the keys of the map should be the operator types, because one `PlatformConfig` struct already represents one operator type.
I think the keys of the map should be the configuration properties. For example:
```
dbaasv1alpha1.ObservabilityInstallation: {
		Name:           OBSERVABILITY_NAME,
		CSV:            OBSERVABILITY_CSV,
		DeploymentName: OBSERVABILITY_DEPLOYMENT,
		Image:          OBSERVABILITY_CATALOG_IMG,
		PackageName:    OBSERVABILITY_PKG,
		Channel:        OBSERVABILITY_CHANNEL,
		DisplayName:    OBSERVABILITY_DISPLAYNAME,
		Type:           dbaasv1alpha1.TypeOperator,
		OtherConfigs: map[string]string{
			AuthType:           os.Getenv("AUTH_TYPE"),
			RemoteWritesURL:    os.Getenv("RHOBS_API_URL"),
			RHSSOTokenURL:      os.Getenv("RH_SSO_TOKEN_ENDPOINT"),
			ObservatoriumToken: os.Getenv("RHOBS_TOKEN"),
			AddonName:          os.Getenv("ADDON_NAME"),
			RHOBSSecretName:    fmt.Sprintf("%v-prom-remote-write", os.Getenv("ADDON_NAME")),
		},
	},
```

xieshenzh (changes requested):

api/v1alpha1/dbaasplatform_types.go
redhatHameed: The difference is I am using `map[string]struct{}` which I think is better as we are restricting the params list then  `map[string]string{}`, and it's easy to validate and consume by function like [here](https://github.com/RHEcosystemAppEng/dbaas-operator/blob/fa16d23759df55349e1e7b79a77e440ee79239cb/controllers/reconcilers/providers_installation/observability-operator.go#L66)

xieshenzh: The code of `OtherConfig` actually restricts all the `PlatformConfig` to use only the params list of the oberservability operator, which would still make `OtherConfigs` specific to the oberservability operator.
For example, if I want to use `OtherConfig` for the cockroach operator, I cannot get rid of the obervability operator params list.
In addition, the `map` will always have one key which is the operator type, so it is unnecessary to use a `map` in this case.
If you want the restrict the behavior of the params list, you could create an interface of `OtherConfig`, and make the obersvability operator params struct implement the interface.

redhatHameed: removed then OtherConfigs from PlatformConfig and created separate function [Review](https://github.com/RHEcosystemAppEng/dbaas-operator/compare/fa16d23759df55349e1e7b79a77e440ee79239cb..ad38aca862440b31e135a4de07092a75cee09a8e)

jeremyary:
hey guys, I believe AH has addressed all initial feedback, PTAL? Thanks!

controllers/reconcilers/constants.go
```diff
@@ -151,6 +152,17 @@ var InstallationPlatforms = map[dbaasv1alpha1.PlatformsName]dbaasv1alpha1.Platfo
...
 }
 
+func GetObservabilityConfig() dbaasv1alpha1.ObservabilityConfig {
+	return dbaasv1alpha1.ObservabilityConfig{
+		AuthType:           os.Getenv("AUTH_TYPE"),
+		RemoteWritesURL:    os.Getenv("RHOBS_API_URL"),
+		RHSSOTokenURL:      os.Getenv("RH_SSO_TOKEN_ENDPOINT"),
+		ObservatoriumToken: os.Getenv("RHOBS_TOKEN"),
+		AddonName:          os.Getenv("ADDON_NAME"),
+		RHOBSSecretName:    fmt.Sprintf("%v-prom-remote-write", os.Getenv("ADDON_NAME")),
```
tchughesiv, line 162: what happens if one of these returns empty?

redhatHameed:
@jianrongzhang89 @jeremyary @tchughesiv @xieshenzh  rebased the PR with main- Please have a another look thanks

controllers/reconcilers/constants.go
redhatHameed: it will ignore remote writing part in the obseravility CR like [1 ](https://github.com/redhatHameed/dbaas-operator/blob/c27c438ac1759cc71ad599320f17eaa0923fdc0e/controllers/reconcilers/providersinstallation/observability-operator.go#L142)& [2](https://github.com/redhatHameed/dbaas-operator/blob/c27c438ac1759cc71ad599320f17eaa0923fdc0e/controllers/reconcilers/providersinstallation/observability-operator.go#L156)

tchughesiv (approved):
lgtm

config/default/manager-env-images.yaml
```diff
@@ -19,9 +19,9 @@ spec:
         - name: CSV_VERSION_CONSOLE_TELEMETRY_PLUGIN
           value: console-telemetry-plugin:0.1.4
         - name: RELATED_IMAGE_OBSERVABILITY_CATALOG
-          value: quay.io/rhobs/observability-operator-catalog:0.0.12
+          value: quay.io/rhobs/observability-operator-catalog:0.0.13
         - name: CSV_VERSION_OBSERVABILITY
-          value: observability-operator.v0.0.12
+          value: observability-operator.v0.0.13
```
tchughesiv, line 24: was `make generate bundle` run after these changes? i don't see them reflected in the bundle csv.

tchughesiv (changes requested):

config/default/manager-env-images.yaml
redhatHameed: fixed that @tchughesiv thanks, reflected in CSV file

tchughesiv (dismissed):

jeremyary (approved):

PR #227, opened 2022-08-30
DBAAS-813: allow Connection to use Instance reference instead of Instance ID

xieshenzh:
## Description

GitOps adminstrator wants to reference `DBaaSInstance` within `DBaaSConnection`.

Added a reference to `DBaaSInstance` in the spec of `DBaaSConnection`, as the alternative to `InstanceID`.
DBaaS operator will get the instance id by reading the `DBInstance` object, and then set the `InstanceID` in the spec of the `DBaaSConnection` object.

Either `InstanceID` or `DBInstance` reference must be set when creating a `DBaaSConnection` object.
If `InstanceID` is set in the spec, the `DBInstance` reference will be ignored.

`DBaaSInstance` and `InstanceID` are immutable in the spec of `DBaaSConnection`. 
But if `InstanceID` is not set when creating the `DBaaSConnection` object, it can be set once, so that DBaaS operator can set it. As soon as it is set, it cannot be changed any more.

Jira: https://issues.redhat.com/browse/DBAAS-813

## Verification Steps

Scenario 1: create Connection with Instance reference.
Scenario 2: create Connection with both Instance ID and Instance reference.
Scenario 3: create Connection without Instance ID or Instance reference.
Scenario 4: create Connection with Instance reference, try to change the Instance reference.
Scenario 5: create Connection with Instance ID, try to change the Instance ID.

## Type of change

- [ ] Code Review completed


controllers/dbaasconnection_controller.go
```diff
@@ -92,6 +94,35 @@ func (r *DBaaSConnectionReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		SetConnectionMetrics(inventory.Spec.ProviderRef.Name, inventory.Name, connection, execution)
 		return ctrl.Result{}, nil
 	} else {
+		// Ignore InstanceRef if InstanceID is present
+		if len(connection.Spec.InstanceID) == 0 {
+			if err := r.getInstanceID(ctx, &connection); err != nil {
```
jianrongzhang89, line 99: This solution updates DBaaSConnection's Spec with the derived instanceID, so provider operators are not impacted. This is nice, however, it may have unwanted side effects since it alters the CR's spec, which is typically viewed as the source of the truth. If users use a gitops or other operators to deploy the CR with a helm chart, those operators may mistakenly think that the DBaaSConnection object deployed in a cluster is out of sync with the source, and hence try to reconcile it by redeploying the CR to the cluster, leading to race conditions with the DBaaS operator.
Could this solution be enhanced without touching the original CR spec while keeping provider operators unchanged, ie, retrieve the instanceID and pass it to provider CR directly?

xieshenzh: There won't be race condition with the DBaaS operator, because `InstanceID` and `InstanceRef` are immutable. Any update to the CR would be rejected by the DBaaS webhook.

The reason for setting the `InstanceID` in the DBaaSConnection's spec is to keep it immutable and consistent with the providers. Because the `InstanceID` of the `DBaaSInstance` reference could change, and the `InstanceID` in the provider CR's spec could be changed too. But not all our provider operators can deal with instance change to the Connection.  

If this solution is not acceptable, I think it is better to change the provider operators. Because not every provider operator needs the `InstanceID` to retrieve the connection information. Meanwhile, all the provider operators should be reviewed whether they could handle instance change.

jianrongzhang89: >> "There won't be race condition with the DBaaS operator, because InstanceID and InstanceRef are immutable. Any update to the CR would be rejected by the DBaaS webhook." 
This can cause problems for other operators (such as gitops operator) to fail to deploy the CR, because they consider the CR spec defined in the repo as the source of truth, and it will try to reconcile the objects in the cluster when they detect any difference between the manifest and the objects in the cluster. With the above approach, the deployment will always fail or be out-of-sync.

We do not have to change the provider operator code, and but can simply set the instanceID when the dbaas operator creates the provider CR, see the poc code below:
https://github.com/jianrongzhang89/dbaas-operator/blob/test-gitops/controllers/dbaasconnection_controller.go#L129

Please ping me if you would like to discuss this further. Thanks.

xieshenzh: After the discussion, I'll update the provider operators to handle instance changes.

jeremyary: Why is the preference to handle this in the provider operators instead of deriving the same value we pass today from the reference and use this as the value we pass? We certainly won't update our providers with any further features for this release. I see your point on the ID changing, but my preference would be to minimally implement here and revisit should all these gitops theoreticals become real issues. Right now, it feels like we're trying to cover every what-if scenario and overly complicating this.

api/v1alpha1/dbaasconnection_webhook.go
```diff
@@ -58,6 +58,16 @@ func (r *DBaaSConnection) ValidateDelete() error {
 	return nil
 }
 
+func (r *DBaaSConnection) validateCreateDBaaSConnectionSpec() error {
+	if len(r.Spec.InstanceID) > 0 && r.Spec.InstanceRef != nil && len(r.Spec.InstanceRef.Name) > 0 {
```
jianrongzhang89, line 62: I think we could allow both values here but if instanceID is provided, the instanceRef simply gets ignored. 

In the reconciliation, the operator could issue a warning in the log if the instanceID derived from instanceRef is different from the instanceID from the spec - but this is nice to have for this release I think,

api/v1alpha1/dbaasconnection_webhook.go
```diff
@@ -58,6 +58,16 @@ func (r *DBaaSConnection) ValidateDelete() error {
...
+		return field.Invalid(field.NewPath("spec").Child("instanceID"), r.Spec.InstanceID, "both instanceID and instanceRef are specified")
+	}
+	if len(r.Spec.InstanceID) == 0 && (r.Spec.InstanceRef == nil || len(r.Spec.InstanceRef.Name) == 0) {
+		return field.Invalid(field.NewPath("spec").Child("instanceID"), r.Spec.InstanceID, "both instanceID and instanceRef are not specified")
```
jianrongzhang89, line 66: both instanceID and instanceRef are not specified => either instanceID or instanceRef should be specified

controllers/dbaasconnection_controller.go
```diff
@@ -219,3 +221,41 @@ func (r *DBaaSConnectionReconciler) Delete(e event.DeleteEvent) error {
...
+		return spec
+	}
+
+	if !reflect.DeepEqual(instance.Spec.InventoryRef, spec.InventoryRef) {
+		logger.Error(fmt.Errorf("instance and connection don't use the same inventory reference"), "Cannot use the instance reference")
+		return spec
+	}
+
+	if len(instance.Status.InstanceID) == 0 {
+		logger.Error(fmt.Errorf("instance ID is not avaiable"), "Cannot use the instance reference")
```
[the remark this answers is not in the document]
xieshenzh: Users can create connections with instance id when the instances are not ready. So the check is not necessary.

api/v1alpha1/dbaasconnection_webhook.go
xieshenzh: It seems overly complicating the feature.

jianrongzhang89: What I mean was to remove this checking, making it simpler.

controllers/dbaasconnection_controller.go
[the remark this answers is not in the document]
jianrongzhang89: If the instance provisioning is still on going and the application tries to connect to the db, the application may fail. We need to test such cases out, but since this is the existing behavior, I am fine to keep it as it is.

jeremyary (approved):

PR #233, opened 2022-09-14
[DBAAS-887] add selector option for valid DBaaS Connection/Instance namespaces

tchughesiv:

## Description
Support for setting namespace [label selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/). The following can be set in policy and/or inventory objects.
```yaml
spec:
  connectionNsSelector:
    matchExpressions:
      - key: test
        operator: NotIn
        values:
          - production
    matchLabels:
    - foo: one
    - two: bar
```

### DIFF
--- a/api/v1alpha1/dbaaspolicy_types.go
+++ b/api/v1alpha1/dbaaspolicy_types.go
@@ -34,7 +34,13 @@ type DBaaSInventoryPolicy struct {
 	// Namespaces where DBaaSConnections/DBaaSInstances are allowed to reference a policy's inventories.
 	// Each inventory can individually override this. Use "*" to allow all namespaces.
 	// If not set in either the policy or inventory object, connections will only be allowed in the inventory's namespace.
-	ConnectionNamespaces []string `json:"connectionNamespaces,omitempty"`
+	ConnectionNamespaces *[]string `json:"connectionNamespaces,omitempty"`
+
+	// Use a label selector to determine namespaces where DBaaSConnections/DBaaSInstances are allowed to reference a policy's inventories.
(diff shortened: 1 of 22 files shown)


api/v1alpha1/dbaaspolicy_types.go
```diff
@@ -34,7 +34,11 @@ type DBaaSInventoryPolicy struct {
 	// Namespaces where DBaaSConnections/DBaaSInstances are allowed to reference a policy's inventories.
 	// Each inventory can individually override this. Use "*" to allow all namespaces.
 	// If not set in either the policy or inventory object, connections will only be allowed in the inventory's namespace.
-	ConnectionNamespaces []string `json:"connectionNamespaces,omitempty"`
+	ConnectionNamespaces *[]string `json:"connectionNamespaces,omitempty"`
+
+	// Use label selectors to determine namespaces where DBaaSConnections/DBaaSInstances are allowed to reference a policy's inventories.
```
raffaelespazzoli, line 39: this is not correct: a label selector is not a map[string]string, the following type should be used: `metav1.LabelSelector` is has utility functions to decide whether something is selected and support expressions.

raffaelespazzoli (changes requested):

api/v1alpha1/dbaaspolicy_types.go
tchughesiv: tks for the feedback, agreed that makes more sense.

tchughesiv: @raffaelespazzoli better?

jeremyary (approved):

PR #260, opened 2022-10-11
[DBAAS-940] add ack dependency to ensure seamless upgrade

tchughesiv:


redhatHameed:
@tchughesiv do we need JIRA or we can add PR link in existing one

redhatHameed:
> @redhatHameed i think we need separate one right?

Yes that will be good Thanks

PR #284, opened 2022-10-27
checking in the backstage configuration for testing.

lokeshrangineni:
Checking in the backstage configuration for testing.

### DIFF
--- a/backstage-config/catalog-info.yaml
+++ b/backstage-config/catalog-info.yaml
@@ -1,0 +1,23 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: rhoda-dbaas-operator
+  description: |
+    Rhoda Dbaas operator.
+  links:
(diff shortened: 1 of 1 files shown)


lokeshrangineni:
@jeremyary Not sure what that means but if you are asking to have a discussion about this effort then sure. We can have a call. Looks like you are not aware of this change so some context for you - @ccustine asked me to add a backstage application configuration for this project. I can test on some personal repo before actually making changes to this repo.

bmozaffa:
Interesting, I did not realize Backstage would require adding a config file to the codebase.

In short, backstage is a developer portal. It should provide a one-stop shop for project devs to see where things are, and allow some actions (deploy/provision/etc but unclear) to be done from the portal. Our primary goal is to better understand backstage, but perhaps it will prove useful to the RHODA project too.

jeremyary (approved):

PR #286, opened 2022-10-31
[DBAAS-980] enforce use of properly versioned sdk and opm binaries

tchughesiv:
## Description
enforce use of properly versioned sdk and opm binaries

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,8 @@ CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
 ENVTEST ?= $(LOCALBIN)/setup-envtest
 
 ## Tool Versions
+OPM_VERSION ?= v1.23.2
+OPERATOR_SDK_VERSION ?= v1.20.1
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -272,37 +274,29 @@ bundle-push: ## Push the bundle image.
 	$(MAKE) docker-push IMG=$(BUNDLE_IMG)
 
 .PHONY: opm
-OPM = ./bin/opm
+OPM = ./bin/opm.$(OPM_VERSION)
```
jianrongzhang89, line 277: Why this change is needed - to avoid using an outdated opm under the bin folder? It is downloaded during the build.

tchughesiv: Yes, that is one reason for the change... but also to avoid using the wrong version in a user's workstation PATH. Note the other changes made to this make command. With this change, we'd ONLY use the properly versioned `opm` within the repo's relative ./bin dir.

jeremyary (approved):

PR #294, opened 2022-11-23
DBAAS-999 Full Provisioning with CockroachDB Provider - data model changes

jianrongzhang89:
## Description
Implements the data model changes for dbaas operator
See ADR: https://docs.google.com/document/d/1IEgyNOdryWR17cj_QtMsa62ZZtk0Hqdst-NtGGn9EFg/edit#

Reviewers: If you have any questions/comments about the solution itself, please comment in the ADR.  For implementation/coding issues please comment on this PR. Thank you.

## Type of change

- [X ] New feature (non-breaking change which adds functionality)

- [X ] Code Review completed


api/v1alpha1/dbaasprovider.go
```diff
@@ -288,6 +299,15 @@ type DBaaSProviderInstance struct {
 	Status DBaaSInstanceStatus `json:"status,omitempty"`
 }
 
+// ParamData defines a key/value pair for parameters
+type ParamData struct {
```
bmozaffa, line 303: Do we use / care about this data structure, or is it just metadata we ignore? If the former, please use strict definitions that make sense for a given context

api/v1alpha1/dbaasprovider.go
```diff
@@ -302,6 +322,32 @@ type InstanceParameterSpec struct {
 	// Define if this field is required or not.
 	Required bool `json:"required"`
 
-	// Default value for this field.
+	// Depends dependent parameters. Each dependent parameter must be registered
+	Depends []ParamData `json:"depends,omitempty"`
```
bmozaffa, line 326: Dependent parameter? Can you elaborate?

bmozaffa (changes requested):

api/v1alpha1/dbaasprovider.go
jianrongzhang89: These are other parameters (such as pricing plan and cloud provider) that must be set for this field to apply and be visible to the user.

api/v1alpha1/dbaasprovider.go
jianrongzhang89: Yes this is used. I renamed it to ParameterData.

api/v1alpha1/dbaasprovider.go
bmozaffa: ah ok, so maybe `FieldDependency`?

api/v1alpha1/dbaasprovider.go
bmozaffa: That still doesn't tell me, as a reader / reviewer, what this is and what it means. The type, as well as its fields, should have names that relate to DBaaS providers.

jianrongzhang89: @bmozaffa thanks! You are right. The meaning was not clear (I wanted to use the same type for two purposes). I switched to use two types instead:
type DropdownItem struct {
	// Identifier for the value
	Key string `json:"key,omitempty"`

	// Corresponding display value
	Value string `json:"value,omitempty"`
}

// FieldDependency defines the name and value of a field used as a dependency
type FieldDependency struct {
	// Name of the field used as a dependency
	Name string `json:"name,omitempty"`

	// Value of the field used as a dependency
	Value string `json:"value,omitempty"`
}

Let me know what you think.

api/v1alpha1/dbaasprovider.go
jianrongzhang89: Sounds good. Thanks.

api/v1alpha1/dbaasprovider.go
```diff
@@ -288,6 +299,24 @@ type DBaaSProviderInstance struct {
 	Status DBaaSInstanceStatus `json:"status,omitempty"`
 }
 
+// DropdownItem defines a key/value pair for a dropdown item
+type DropdownItem struct {
+	// Identifier for the value
+	Key string `json:"key,omitempty"`
+
+	// Corresponding display value
+	Value string `json:"value,omitempty"`
```
bmozaffa, line 308: What's key and what's value?

api/v1alpha1/dbaasprovider.go
```diff
@@ -288,6 +299,24 @@ type DBaaSProviderInstance struct {
...
+}
+
+// FieldDependency defines the name and value of a field used as a dependency
+type FieldDependency struct {
+	// Name of the field used as a dependency
+	Name string `json:"name,omitempty"`
+
+	// Value of the field used as a dependency
+	Value string `json:"value,omitempty"`
```
bmozaffa, line 317: What's name and what's value?

bmozaffa (changes requested):

api/v1alpha1/dbaasprovider.go
jianrongzhang89: Key is the index for the dropdown item
Value is the display value of the item in the dropdown.

api/v1alpha1/dbaasprovider.go
jianrongzhang89: Name: this is the name of the field used as a dependency, for instance, "cloud_provider".
Value: the value used as a dependency, ie, "AWS" -- this means this dropdown list is only applicable to cloud_provider = "AWS".

bmozaffa: Ok, so use something like FieldName and FieldValue or simply Field and Value instead of Name/Value?

api/v1alpha1/dbaasprovider.go
bmozaffa: Index, as in the order in which it should be sorted? Aren't arrays already sorted? Or if necessary, just call it Index?

bmozaffa (changes requested):

api/v1alpha1/dbaasprovider.go
jianrongzhang89: Yes the array is sorted. However, we'd better use the key-value pairs (for instance "AWS" for "Amazon Web Services") than use an integer index. So I prefer to keep it as "key" here.

api/v1alpha1/dbaasprovider.go
jianrongzhang89: Field/ value sounds good to me :)

api/v1alpha1/dbaasprovider.go
bmozaffa: Ok, I was not clear on what you meant by index. Maybe `Value` and `DisplayValue`?

api/v1alpha1/dbaasprovider.go
bmozaffa: Thanks, further discussed elsewhere.

api/v1alpha1/dbaasprovider.go
```diff
@@ -302,6 +331,35 @@ type InstanceParameterSpec struct {
...
+	HelpText string `json:"helpText,omitempty"`
+
+	// Default value for this field
 	DefaultValue string `json:"defaultValue,omitempty"`
 }
+
+// DBaaSProviderData dropdown list data for UX
+type DBaaSProviderData struct {
+	// Parameter name
+	Parameter string `json:"parameter,omitempty"`
```
bmozaffa, line 355: Field? We'll need to make sure we're consistent wherever we end up with this.

jianrongzhang89: OK. Thanks.

api/v1alpha1/dbaasprovider.go
jianrongzhang89: Yeah, Value/DisplayValue looks much better. Thank you for your suggestion.

api/v1alpha1/dbaasprovider.go
```diff
@@ -302,6 +331,35 @@ type InstanceParameterSpec struct {
...
+// DBaaSProviderData dropdown list data for UX
+type DBaaSProviderData struct {
+	// Field name
+	Field string `json:"field,omitempty"`
+
+	// List of the dependent fields and their values
+	Dependencies []FieldDependency `json:"dependencies,omitempty"`
+
+	// Data displayed in the dropdown list
+	Data []DropdownItem `json:"data,omitempty"`
```
bmozaffa, line 361: `Options` instead of `Data`?

api/v1alpha1/dbaasprovider.go
```diff
@@ -110,7 +113,15 @@ type DBaaSProviderSpec struct {
 	InstanceParameterSpecs []InstanceParameterSpec `json:"instanceParameterSpecs"`
 }
 
-// Defines the information for a DBaaSProvider object.
+// DBaaSProviderStatus defines the observed state of DBaaSProvider
+type DBaaSProviderStatus struct {
```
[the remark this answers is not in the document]
jianrongzhang89: Yes the provider operator provides the dbaasprovider data as a CR (we decided in the beginning of RHODA project to use a CR instead of a configmap). The reason that why we set the status here, is that some of the data can change from time to time (such as available regions to create new clusters) and also since such data can only be retrieved using necessary credentials through the provider accounts, here we use the inventory's controller to "reconcile" such CR and provide the result in the status.

api/v1alpha1/dbaasprovider.go
jianrongzhang89: "Options" sounds good.

api/v1alpha1/dbaasprovider.go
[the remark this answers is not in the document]
bmozaffa: The inventory's controller? Do you mean the provider operator or something else?

Let's think of the Spec/Status as the request/response. Is this data updated by someone other than the party that initially created the CR and filled in the Spec?

api/v1alpha1/dbaasprovider.go
```diff
@@ -302,6 +331,35 @@ type InstanceParameterSpec struct {
...
+
+	// Additional info about the field
+	HelpText string `json:"helpText,omitempty"`
+
+	// Default value for this field
+	DefaultValue string `json:"defaultValue,omitempty"`
+}
+
+// DBaaSProviderData dropdown list data for UX
+type DBaaSProviderData struct {
```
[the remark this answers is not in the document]
jianrongzhang89: What about "FieldOptions"? This structure now is for the list of dropdown options for a field. What aboutnot to have "Provisioning" in the name so that later it might be used for non-provisioning purposes? Please let me know.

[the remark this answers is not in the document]
jianrongzhang89: In addition, I changed DropdownItem to "Option" which is more generic and can be used for different controls such as dropdown, radio button and checkbox. Please let me know what you think.

api/v1alpha1/dbaasprovider.go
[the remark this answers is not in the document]
jianrongzhang89: This is the provider operator's inventory reconciler. So it is created and controlled by the provider operator only. Currently, the provider operator creates this CR upon start-up so that the operator gets registered with the DBaaS Operator. However, at this point, the provider operator does not have the credentials needed to populate certain data.   Such credentials only become available when a user creates an inventory, and that's why such data population is deferred to the inventory's reconciler and stored in the status of the CR.

The alternative solution would require the provider operator to use a generic "service account" to access the ISV, so that the dbaasprovider CR is managed by a single reconciled. However, this would make our overall solution a lot more complicated due to the need to maintain such a service account.

Please let me know what you think about the approach proposed.

api/v1alpha1/dbaasprovider.go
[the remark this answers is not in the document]
bmozaffa: > What about "FieldOptions"? This structure now is for the list of dropdown options for a field. What aboutnot to have "Provisioning" in the name so that later it might be used for non-provisioning purposes? Please let me know.

It makes sense to not limit this to provisioning... I think DBaaSProviderData is more meaningful than FieldOptions.

api/v1alpha1/dbaasprovider.go
```diff
@@ -110,7 +113,15 @@ type DBaaSProviderSpec struct {
 	InstanceParameterSpecs []InstanceParameterSpec `json:"instanceParameterSpecs"`
 }
 
-// Defines the information for a DBaaSProvider object.
+// DBaaSProviderStatus defines the observed state of DBaaSProvider
+type DBaaSProviderStatus struct {
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// FieldOptions used for dropdown lists in UX
+	FieldOptions []FieldOptions `json:"fieldOptions,omitempty"`
```
bmozaffa, line 121: Should this field (versus the type) then be called ProvisioningFields or similar?

api/v1alpha1/dbaasprovider.go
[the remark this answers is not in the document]
bmozaffa: Service account approach is not possible.

Ok, so what you're saying is that the only difference between initial provider data currently in Spec versus later data currently in status is that the later data was not available to the operator when it was registering itself. That's not a good reason to move that data to Status. It should still be in Spec.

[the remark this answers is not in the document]
jianrongzhang89: There are benefits of using the status to store such data - it will allow the operator to set the status conditions to indicate whether this data has been processed successfully, and in the case when it fails to process it, set the error messages accordingly. In addition, this allows the operator to keep track of when the data was processed last time, and decide whether it needs to refresh it.

Normally the spec should only indicate the desired state of the CR, so the spec is static and not good for storing data that can change later.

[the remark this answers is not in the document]
bmozaffa: Good points, the whole thing is a bit weird, but ok!

api/v1alpha1/dbaasprovider.go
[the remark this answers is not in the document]
jianrongzhang89: OK. Changed back to DBaaSProviderData.

api/v1alpha1/dbaasprovider.go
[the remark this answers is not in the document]
jianrongzhang89: @xieshenzh I agree this is not a perfect solution, but it would meet the requirements for CockroachDB. If in the future we indeed need to support the case where different credentials would result in a different list of regions we could enhance it for that provider (ie, store such special data in the inventory status) and it can be considered as out of scope for the current task.

About the security concern - this is what we have today and this design does not make it worse. Since the CR is a cluster-level custom resource, users other than the provider operator's service account will need a cluster admin role to modify the data.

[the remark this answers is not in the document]
xieshenzh: It will meet the requirements for CockroachDB today, but CockroachDB could change the retrieved data whenever they like.
And I don't think any users with the permission to read the DBaaSProvider CR should have access to the tenants' data by default.
Are there any important reasons for which we should not store such data in the inventory status now? I think it would be more difficult to move it to the inventory status later.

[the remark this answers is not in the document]
bmozaffa: Maybe I made some wrong assumptions in my read of this... I assumed the fact that you could only retrieve field values using credentials was a fluke, in that this should be publicly available data and also that it is global in context. Is that not the case? Can we ask cockroach?

If the data is really different per tenant, that is a problem even today, because multiple inventories are supported for cockroach.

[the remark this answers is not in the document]
jianrongzhang89: @bmozaffa yes we can follow up with Cockroach for clarification.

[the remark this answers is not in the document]
xieshenzh: It sounds like they are suggesting using static values for the available regions.
I guess that is what they are doing with their own UI.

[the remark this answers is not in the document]
jianrongzhang89: What they suggest is that it is fine to cache the data for a certain period of time, but since the data can still change, it is better to have the ability to refresh data than to require restarting of the provider operator itself.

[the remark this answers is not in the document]
bmozaffa: Let's ask and get an idea of the scale when they say "slowly"... are they adding a new region every 3-4 months? Any other changes, and anything more frequently?

api/v1alpha1/dbaasprovider.go
jianrongzhang89: Changed to "Options".

api/v1alpha1/dbaasprovider.go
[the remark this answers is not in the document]
jianrongzhang89: @bmozaffa This what they replied:
"Regions will not be removed from the list. New additions are made irregularly so it's tough to give an indication of how often that might happen - refreshing the cache every week seems reasonable - so your customers would go no longer than 1 week without knowledge of a new region having become available. One big event to keep an eye out for, especially from an ARO perspective, is the addition of Azure regions. This is planned for May/June 2023 timeframe but could come online sooner. "

redhatHameed:
@jianrongzhang89 we can we add new one or update existing sample example for provider registration under -WDYT ?>https://github.com/RHEcosystemAppEng/dbaas-operator/blob/main/config/samples/dbaas_v1alpha1_dbaasprovider.yaml

api/v1alpha1/dbaasprovider.go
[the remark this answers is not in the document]
xieshenzh: If you insist on storing the data in the dbaas provider object, then why not store the data in the spec of the dbaas provider object?
Just modify `InstanceParameterSpec` to store the candidate values, and use the crdb inventory controller to update the spec of the dbaas provider object, it is not immutable. The crdb provider operator creates the dbaas provider object, it can also update the spec of the dbaas provider object. 
It doesn't make sense to set the status of the dbaas provider object with the crdb inventory controller.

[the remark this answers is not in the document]
jianrongzhang89: @xieshenzh I agree the provider data for the dropdown list is better included in the dbaasprovider CR spec instead of the status.

api/v1alpha1/dbaasprovider.go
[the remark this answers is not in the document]
bmozaffa: With the changes we discussed, we should now move away from the abstract approach and make this very specific to our provisioning page. So should this be called ProvisioningParameter (each is a single param)?

api/v1alpha1/dbaasprovider.go
```diff
@@ -302,6 +328,24 @@ type InstanceParameterSpec struct {
...
+	HelpText string `json:"helpText,omitempty"`
+
+	// Default value for this field
 	DefaultValue string `json:"defaultValue,omitempty"`
 }
+
+// DBaaSProviderData defines options available for UX for a field
+type DBaaSProviderData struct {
+	// Field name
+	Field string `json:"field,omitempty"`
```
bmozaffa, line 341: Create enum for this, representing the fields we have for provisioning

api/v1alpha1/dbaasprovider.go
```diff
@@ -288,12 +296,30 @@ type DBaaSProviderInstance struct {
...
+	Value string `json:"value,omitempty"`
+
+	// Corresponding display value
+	DisplayValue string `json:"displayValue,omitempty"`
+}
+
+// FieldDependency defines the name and value of a field used as a dependency
+type FieldDependency struct {
+	// Name of the field used as a dependency
+	Field string `json:"field,omitempty"`
```
bmozaffa, line 311: Create enum for this, representing the fields we have for provisioning

bmozaffa (changes requested):

jianrongzhang89:
> 

Sounds good. Will add one once our design is finalized.

api/v1alpha1/dbaasprovider.go
jianrongzhang89: @bmozaffa currently, different providers already use their own list of parameters (for freetrial provisioning). If we create enum to enforce the list of parameters, we may run into issues when we upgrade to the new RHODA release, as it is possible there are existing CRs with old parameter names defined, and the upgrade will fail unless a data migration is performed. Without the enums, the provider operators could then map both the old and new parameters to the desired one to allow a smooth upgrade.

On the other hand, since for R6 we already plan CRD versioning upgrade, we could add the above migration to the CRD upgrade using the conversion webhook. 

How should we proceed? @jeremyary

api/v1alpha1/dbaasprovider.go
jianrongzhang89: Added.Please review.

api/v1alpha1/dbaasprovider.go
jianrongzhang89: Added the enum. Please review.

bmozaffa: I see the kube builder annotation for enum... but why not an enum in Go?

api/v1alpha1/dbaasprovider.go
```diff
@@ -288,12 +296,32 @@ type DBaaSProviderInstance struct {
...
+
+	// Value of the field used as a dependency
+	Value string `json:"value,omitempty"`
+}
+
+// Indicates what fields to collect from the user interface, and how to display those fields in a form to provision a database instance.
 type InstanceParameterSpec struct {
+	// +kubebuilder:validation:Enum=name;plan;cloud_provider;regions;availability_zones;nodes;machine_type;storage_gib;spend_limit;project
```
bmozaffa, line 320: I don't expect you would need this... as we discussed on the call, we are fully moving away from this abstract model of the provider specifying what fields are needed.

jianrongzhang89: Got it. Updated the PR.

api/v1alpha1/dbaasprovider.go
```diff
@@ -21,6 +21,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type FieldName string
```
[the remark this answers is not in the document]
bmozaffa: Then maybe call the other one something like ProvisioningParamRequirements?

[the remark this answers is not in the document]
jianrongzhang89: @bmozaffa I think it is better to keep the other one as ProvisioningParameter. What about changing this one to "ProvisioningParameterName"?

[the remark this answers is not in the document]
jianrongzhang89: Changing FieldName to ProvisioningParameter sounds good. What about changing the other one to ProvisioningParameterData?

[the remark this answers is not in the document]
bmozaffa: Sure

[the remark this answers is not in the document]
jianrongzhang89: Thanks! PR updated.

api/v1alpha1/dbaasprovider.go
bmozaffa: I'm not sure I understand the migration issues... and ideally we would want to enforce this enum in the CRD as well

api/v1alpha1/dbaasprovider.go
bmozaffa: This still remaining?

jianrongzhang89: No, it is already removed

api/v1alpha1/dbaasprovider.go
jianrongzhang89: We can postpone such CRD validation to a later release if needed. Since the list of parameters and corresponding data is coded in the provider operators rather than provided by end users, the chance if having an invalid parameter can be eliminated as long as we perform the necessary code review and integration tests before release.

api/v1alpha1/dbaasprovider.go
bmozaffa: I don't think we're talking about the same thing. The provider should no longer be providing `InstanceParameterSpecs []InstanceParameterSpec` to specify which fields are needed. We know what fields are needed now.

jianrongzhang89: The InstanceParameterSpec defines which fields are used for the provider, and also provides the display value of for each field. If we do not use it we have to invent something else for such purpose.

bmozaffa: Let's start with looking at lines 124-125 (versus the type itself). The description says `Indicates what parameters to collect from the user interface`. That's no longer a question and we know what parameters should be collected.

jianrongzhang89: We still want to have the list of parameters that are actually used for each provider. Rather than introducing another type with a boolean flag, it would be easier to simply list the parameters used along with the display value needed by the UX in this existing data structure (For some fields such as spend_limit, it can include the default value the UX can set as well). The ones that are not on the list will be ignored by the UX, indicating that such parameters are not applicable to that provider.

bmozaffa: That's not what we discussed on the call. We need to have code that's easier to understand and use. Our provisioning form is now well defined, and most of the fields are in fact required. Providers should only specify if the optional fields apply to them or not.

jianrongzhang89: This is a different approach and it will work, but will likely require more amount of implementation work. This is because both the UX and backend provider operators are already implemented to use existing InstanceParameterSpec.

api/v1alpha1/dbaasprovider.go
```diff
@@ -59,9 +61,22 @@ const (
...
+	ProvisioningName              ProvisioningParameterType = "name"
+	ProvisioningPlan              ProvisioningParameterType = "plan"
+	ProvisioningCloudProvider     ProvisioningParameterType = "cloudProvider"
+	ProvisioningRegions           ProvisioningParameterType = "regions"
+	ProvisioningAvailabilityZones ProvisioningParameterType = "availabilityZones"
+	ProvisioningNodes             ProvisioningParameterType = "nodes"
+	ProvisioningMachineType       ProvisioningParameterType = "machineType"
+	ProvisioningStorageGib        ProvisioningParameterType = "storageGib"
+	ProvisioningSpendLimit        ProvisioningParameterType = "spendLimit"
+	ProvisioningProject           ProvisioningParameterType = "project"
```
redhatHameed, line 79: is this specific project to mongoldb ?

api/v1alpha1/dbaasprovider.go
```diff
@@ -288,20 +308,64 @@ type DBaaSProviderInstance struct {
...
+	Name              *ProvisioningParameter `json:"name,omitempty"`
+	Plan              *ProvisioningParameter `json:"plan,omitempty"`
+	CloudProvider     *ProvisioningParameter `json:"cloudProvider,omitempty"`
+	Regions           *ProvisioningParameter `json:"regions,omitempty"`
+	AvailabilityZones *ProvisioningParameter `json:"availabilityZones,omitempty"`
+	Nodes             *ProvisioningParameter `json:"nodes,omitempty"`
+	MachineType       *ProvisioningParameter `json:"machineType,omitempty"`
+	StorageGib        *ProvisioningParameter `json:"storageGiB,omitempty"`
+	SpendLimit        *ProvisioningParameter `json:"spendLimit,omitempty"`
+	Project           *ProvisioningParameter `json:"project,omitempty"`
```
redhatHameed, line 340: is this specific 'project' to mongoldb ?

tchughesiv:
looks good... I've added a `hold` label as this change will need to be rebased onto https://github.com/RHEcosystemAppEng/dbaas-operator/pull/296 once it's merged... and all api changes moved to the `v1beta1` dbaasprovider object.

api/v1alpha1/dbaasprovider.go
```diff
@@ -288,20 +310,66 @@ type DBaaSProviderInstance struct {
...
+	// Additional info about the field
+	HelpText string `json:"helpText,omitempty"`
 
-	// Define if this field is required or not.
-	Required bool `json:"required"`
+	// Lists of options for the field. May have multiple option lists based on the dependecies.
+	OptionsLists []ProvisioningParameterOptions `json:"optionsLists,omitempty"`
 
-	// Default value for this field.
+	// Default value for the field
```
bmozaffa, line 358: Is default value redundant, since there is already a default option in the options list?

api/v1alpha1/dbaasprovider.go
```diff
@@ -288,20 +310,66 @@ type DBaaSProviderInstance struct {
...
+	Regions           *ProvisioningParameter `json:"regions,omitempty"`
+	AvailabilityZones *ProvisioningParameter `json:"availabilityZones,omitempty"`
+	Nodes             *ProvisioningParameter `json:"nodes,omitempty"`
+	MachineType       *ProvisioningParameter `json:"machineType,omitempty"`
+	StorageGib        *ProvisioningParameter `json:"storageGiB,omitempty"`
+	SpendLimit        *ProvisioningParameter `json:"spendLimit,omitempty"`
+	Project           *ProvisioningParameter `json:"project,omitempty"`
+	LocationLabel     *ProvisioningParameter `json:"locationLabel,omitempty"`
+	HardwareLabel     *ProvisioningParameter `json:"hardwareLabel,omitempty"`
+}
```
bmozaffa, line 345: Somewhat of an advanced Go design question for anyone with time to dig into it, @jianrongzhang89  @xieshenzh @tchughesiv , but I wonder if there is a way to not repeat these names, once for the enum and once for the list of parameters here. Not worth holding up the PR, though!

bmozaffa (changes requested):

api/v1alpha1/dbaasprovider.go
xieshenzh: Do you mean the list of `ProvisioningParameterType` constants?
Maybe, use a real enum like this: https://go.dev/play/p/tozFyBJSJQM

bmozaffa: Thanks, @xieshenzh , I think that's a good hint at what would work, but maybe still needs some work to address the issue? The redundancy is that the field names are defined [once](https://github.com/jianrongzhang89/dbaas-operator/blob/full-provisioning/api/v1alpha1/dbaasprovider.go#L70-L81) for ParamType and then listed again for each for [ProvisioningParameterSpec](https://github.com/jianrongzhang89/dbaas-operator/blob/full-provisioning/api/v1alpha1/dbaasprovider.go#L333-L344).

api/v1alpha1/dbaasprovider.go
jianrongzhang89: The default value is used for parameters that do not have dropdown options, for instance, spend limit. The default option is for fields that are displayed as dropdowns.

api/v1alpha1/dbaasprovider.go
jianrongzhang89: Yes, "project" is for mongo, and it can be used as team_id for crunchy later.

redhatHameed: can we make something like, `Team Project`  so this will work for both WDYT?

api/v1alpha1/dbaasprovider.go
bmozaffa: @tchughesiv can you clarify this for us? I believe if a CR with old version is provided, that will be taken care of, by the conversion webhook, and we don't need to disable validation for this, right?

api/v1alpha1/dbaasprovider.go
jianrongzhang89: @xieshenzh @bmozaffa this sounds good. Updated the PR.

api/v1alpha1/dbaasprovider.go
jianrongzhang89: @bmozaffa now with the new data model,  CRD validation for the list of fields should be fine. I updated the PR accordingly.

api/v1alpha1/dbaasprovider.go
jianrongzhang89: This sounds good to me. Updated.

api/v1alpha1/dbaasprovider.go
bmozaffa: I'm not sure what you're trying to achieve here... my (low priority) concern was that the list of parameters was repeated twice (see my links above). Now it is repeated three times: [here](https://github.com/jianrongzhang89/dbaas-operator/blob/full-provisioning/api/v1alpha1/dbaasprovider.go#L82-L93) and [here](https://github.com/jianrongzhang89/dbaas-operator/blob/full-provisioning/api/v1alpha1/dbaasprovider.go#L337-L348) and [here](https://github.com/jianrongzhang89/dbaas-operator/blob/full-provisioning/api/v1alpha1/dbaasprovider.go#L358-L369)

jianrongzhang89: @bmozaffa my bad. Fixed.

bmozaffa: Thoughts, @xieshenzh ?

xieshenzh: I think the constants are supposed to be used by the provider operators, which improves the usability of the API and the readability of the provider operator code.
If we don't care about the usability or readability, we don't really need the constants.
On the other hand, I am not sure if there are other good options. 
I don't think it is easy to design a function which outputs the field name in the struct, while the input is the field value or reference. If we use the index of field in the struct as the input, the usability and readability will not be good.

Another option is to remove the names in the `ProvisioningParametersSpec` instead of the constants.
The `ProvisioningParametersSpec` is created by the provider operators and used by the UX code internally.
We can change it to be a list a parameters as it used to be, with which we won't need the names in the code.

xieshenzh: There is another option, which is to create a code generator to generate the constants instead of hardcoding them here in the file.

bmozaffa: No, that's fine, don't need to go to those lengths. However, I'm not sure about where we've ended up. @jianrongzhang89 you don't even seem to be using `ProvisioningParameterType` in the current design?

jianrongzhang89: @bmozaffa The latest design does not use ProvisioningParameterType directly. However, the const values defined for ProvisioningParameterType will be used by the provider operators to convert from common fields to operator's own parameters so it is still useful. What do you think?

bmozaffa: Yeah, I think that's missing the point. I want an enum that enforces the design as well as helps the user understand what goes where.

Ok, so assuming we have a simple string enum:
```
const (
	ProvisioningName ProvisioningParamField = "Name"
	ProvisioningPlan = "Plan"
	ProvisioningCloudProvider = "CloudProvider"
	ProvisioningRegions = "Regions"
	ProvisioningAvailabilityZones = "AvailabilityZones"
	ProvisioningNodes = "Nodes"
	ProvisioningMachineType = "MachineType"
	ProvisioningStorageGib = "StorageGib"
	ProvisioningSpendLimit = "SpendLimit"
	ProvisioningTeamProject = "TeamProject"
	ProvisioningLocationLabel = "LocationLabel"
	ProvisioningHardwareLabel = "HardwareLabel"
)
```

Then it definitely needs to be used in field dependency:

```
type FieldDependency struct {
	Field ProvisioningParamField `json:"field,omitempty"`
```

and I understand that it can't be used in our struct definition, but what if we had no struct at all, and since you're making everything optional there anyway, could we have this under DBaaSProviderSpec?

```ProvisioningParametersSpec map[ProvisioningParamField]ProvisioningParametersSpec 'json:"provisioningParametersSpec,omitempty"```

Comment (bmozaffa) [reply api/v1alpha1/dbaasprovider.go]: Thanks, @jianrongzhang89   To be clear, I don't know and haven't tried to understand any downsides of going this route, so let me know what you think, and you too, @xieshenzh

Comment (jianrongzhang89) [reply api/v1alpha1/dbaasprovider.go]: @bmozaffa It sounds like an excellent idea to me :). I updated the PR accordingly. Please take a look again. Meanwhile, I am working on the crdb operator implementation based on this one and see how it goes.

Comment (jianrongzhang89) [reply api/v1alpha1/dbaasprovider.go]: @bmozaffa this is the cockroach CR populated by the CRDB operator:
https://gist.github.com/jianrongzhang89/a44ae85cb872a71295fd00420ef42531

**Review by bmozaffa — APPROVED** [review]

File: api/v1alpha1/dbaasprovider.go [hunk 18]
```diff
@@ -59,11 +59,31 @@ const (
...
 	TypeLabelKey      = "db-operator/type"
 	TypeLabelKeyMongo = "atlas.mongodb.com/type"
 )
 
+type ProvisioningParameterType string
+
+// DBaaS provisioning fields
+const (
+	ProvisioningName              ProvisioningParameterType = "name"
+	ProvisioningPlan              ProvisioningParameterType = "plan"
```
xieshenzh, line 73: `plan` is what we are going to use to represent the database engine type for RDS, right? @jeremyary @jianrongzhang89

bmozaffa: No, and I think we're missing DatabaseType or whatever we want to call it.

jianrongzhang89: @bmozaffa @xieshenzh OK. So use "databaseType"? if so I will add it.

bmozaffa: No strong opinion here, but basically something like that, which would apply to AWS, GCP and Azure services and/or anyone else who isn't doing just a single type of database

jianrongzhang89: @bmozaffa @xieshenzh PR updated. Please take another look.

bmozaffa: Looks good to me!

redhatHameed (dismissed):
lgtm

api/v1alpha1/dbaasprovider.go
```diff
@@ -288,20 +306,63 @@ type DBaaSProviderInstance struct {
...
+	HelpText string `json:"helpText,omitempty"`
 
-	// Define if this field is required or not.
-	Required bool `json:"required"`
+	// Lists of options for the field. May have multiple option lists based on the dependencies.
+	// If options lists are present, the field is displayed as a dropdown in the UX
+	OptionsLists []ProvisioningParameterOptions `json:"optionsLists,omitempty"`
+
+	// Lists of default values for the field. May have multiple default values based on the dependencies.
+	// If default values are present, the field is displayed as a text input field.
```
bmozaffa, line 341: Instead of duplicating the structure for text input field, how about you rename ProvisioningParameterOptions to something like ConditionalProvisioningParameters, and rename DefaultOption to DefaultValue? Then if `Options []Option` is not provided, it's an input field. If it is provided, DefaultValue is the default drop down option, and its DisplayValue is already known from the array anyway.

Still leave the ambiguity, but that's the same as before/now, and I think ok.

bmozaffa (changes requested):

api/v1alpha1/dbaasprovider.go
jianrongzhang89: I used the DefaultOption instead of DefaultValue as requested by @olavtar for the simplicity of the UX implementation so that the UX can set the defaultOption in the dropdown without calculating the defaultvalue based on its key first. @olavtar Are you OK with what @bmozaffa suggested above?

olavtar: @jianrongzhang89 sure, we can try that.

jianrongzhang89: @bmozaffa updated based on your suggestion.

bmozaffa (approved):

olavtar (approved):

jianrongzhang89:
This PR is replaced by https://github.com/RHEcosystemAppEng/dbaas-operator/pull/308

PR #306, opened 2023-01-11
add v1beta1 conversion webhook for dbaasplatform api, revert controller changes

tchughesiv:

## Type of change

- [ ] Code Review completed


tchughesiv:
don't merge yet, i need to do a lot of testing first

api/v1alpha1/dbaasplatform_conversion.go
```diff
@@ -0,0 +1,70 @@
...
+*/
+
+package v1alpha1
+
+import (
+	"github.com/RHEcosystemAppEng/dbaas-operator/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/conversion"
+)
+
+// notes on writing good spokes https://book.kubebuilder.io/multiversion-tutorial/conversion.html
```
jeremyary, line 24: Never heard hub/spoke in this context before, had to look back a bit and figure out what they're on about. Makes sense!

jeremyary (commented):
concept/syntax-wise, lgtm. I like the migration out of main into setups, def feels appropriate to take advantage of that phase the same as you do for setting up filtering and what not. Adding to the "lessons we've learned with experience" list :)

tchughesiv:
@redhatHameed some of the errors i'm seeing during an operator upgrade trying to revert to the old method. we may have to keep things as-is if i can't find a way forward - 
```
2023-01-13T16:47:13.248Z INFO controller-runtime.builder Conversion webhook enabled {"GVK": "dbaas.redhat.com/v1beta1, Kind=DBaaSProvider"}
2023-01-13T16:47:16.305Z ERROR setup unable to create controller {"controller": "DBaaSPlatform", "error": "could not get a list of dbaas platform intallation CR: conversion webhook for dbaas.redhat.com/v1alpha1, Kind=DBaaSPlatform failed: the server could not find the requested resource"}
main.main
/opt/app-root/src/main.go:211
runtime.main
/usr/lib/golang/src/runtime/proc.go:255
```

tchughesiv:
@redhatHameed it doesn't appear we can implement conversion webhook for dbaasplatform api without major issue... BUT, i was able to revert the controller changes as you requested. The good news is that this object shouldn't require conversion logic as this operator is the only consumer of that api.

redhatHameed (approved):

redhatHameed:
> @redhatHameed it doesn't appear we can implement conversion webhook for dbaasplatform api without major issue... BUT, i was able to revert the controller changes as you requested. The good news is that this object shouldn't require conversion logic as this operator is the only consumer of that api.

@tchughesiv Thanks and agreed and cr creates only once so we are fine with this , and can merge this if no more worked required.

PR #305, opened 2023-01-11
DBAAS-999 Full Provisioning with CockroachDB Provider - data model changes and v1beta1 support

jianrongzhang89:
## Description
This PR includes the data model changes for full provisioning of CRDB and RDS clusters in PR (https://github.com/RHEcosystemAppEng/dbaas-operator/pull/294) and adoption of v1beta1 APIs with the following additional changes:
1) Add a flag DBaaSAPIVersion to DBaaSProviderSpec to indicate the DBaaS API version (v1alpha1 or v1beta1) the provider supports.
2) Change DBaaS Inventory, Instance and Connection reconcilers so that the corresponding version (v1alpha1 or v1beta1)  of provider custom resources are created.
3) Unit test cases updated.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA or in issue number have been completed
- [ ] Unit tests added that prove the fix is effective or the feature works 
- [ ] Documentation added for the feature
- [ ] all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer

PR #319, opened 2023-01-23
DBAAS-999 Full Provisioning with CockroachDB Provider - data model changes and support v1beta1 API

jianrongzhang89:
This PR replaces  https://github.com/RHEcosystemAppEng/dbaas-operator/pull/305 which has been reviewed and had rebasing issues.

This PR includes the data model changes for full provisioning of CRDB and RDS clusters in PR (https://github.com/RHEcosystemAppEng/dbaas-operator/pull/294) and adoption of v1beta1 APIs with the following additional changes:

Add a flag GroupVersion to DBaaSProviderSpec to indicate the DBaaS API version (v1alpha1 or v1beta1) the provider supports.
Change DBaaS Inventory, Instance and Connection reconcilers so that the corresponding version (v1alpha1 or v1beta1) of provider custom resources are created.
Unit test cases updated. which had rebasing issues.
This PR includes the data model changes for full provisioning of CRDB and RDS clusters in PR (https://github.com/RHEcosystemAppEng/dbaas-operator/pull/294) and adoption of v1beta1 APIs with the following additional changes:

Add a flag GroupVersion to DBaaSProviderSpec to indicate the DBaaS API version (v1alpha1 or v1beta1) the provider supports.
Change DBaaS Inventory, Instance and Connection reconcilers so that the corresponding version (v1alpha1 or v1beta1) of provider custom resources are created.
Unit test cases updated.

## Type of change

- [ ] Code Review completed


api/v1beta1/dbaasprovider.go
```diff
@@ -85,6 +111,9 @@ type DBaaSProviderSpec struct {
 	// Contains information about database provider and platform.
 	Provider DatabaseProviderInfo `json:"provider"`
 
+	// DBaaS API group version supported by the provider
```
tchughesiv, line 114: should we change this description to something like -
`API group version supported by the provider`

this doesn't have to be a `dbaas.redhat.com` type of apiversion correct? this could be `database.mongodb.io` or something similar? We need be sure to support whatever api `groupversion` the provider choses to set here. 

code like the following might be a start if we don't already support this? 
```go
gv, err := schema.ParseGroupVersion(provider.spec.GroupVersion)
if err != nil {
    return nil, err
}
newInv := unstructured.Unstructured{}
newInv.SetGroupVersionKind(gv.WithKind(provider.spec.InventoryKind))
```
please also add a unit test to cover this alternative groupversion scenario, if one doesn't already exist... i didn't see any though.

tchughesiv (changes requested):

api/v1beta1/dbaasprovider_webhook.go
```diff
@@ -30,4 +31,19 @@ func (r *DBaaSProvider) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
+func (r *DBaaSProvider) GetDBaaSAPIGroupVersion() schema.GroupVersion {
+	if len(r.Spec.GroupVersion) > 0 {
+		groupVersion, err := schema.ParseGroupVersion(r.Spec.GroupVersion)
+		if err == nil {
+			return groupVersion
+		}
+		// If the group version is not valid, default to v1alpha1
+	}
+	// Use v1alpha1 if API version is not defined (for backward compatibility)
+	return schema.GroupVersion{
```
tchughesiv, line 43: would it be possible to just return the v1alpha1 api GroupVersion directly?
```go
          return v1alpha1.GroupVersion
```

api/v1beta1/dbaasprovider_webhook.go
```diff
@@ -30,4 +31,19 @@ func (r *DBaaSProvider) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
+func (r *DBaaSProvider) GetDBaaSAPIGroupVersion() schema.GroupVersion {
+	if len(r.Spec.GroupVersion) > 0 {
+		groupVersion, err := schema.ParseGroupVersion(r.Spec.GroupVersion)
+		if err == nil {
```
tchughesiv, line 37: just some items to discuss... not sure we want to do any of it but worth thinking about:
 - should we let an admin know about this error?
 - instead of making assumptions when this field isn't set, maybe set a default for this a field w/ a CRD default? could also set via conversion webhook logic.

api/v1beta1/dbaasprovider.go
tchughesiv: sure, but can't a provider CR have object components based on our data structures, without using our `dbaas.redhat.com` group or version? Why should force the provider to support objects with redhat in the group name? The CRD GVK should have no bearing on what you've stated.

jianrongzhang89: That was the result of our earlier decision to keep our solution simple. It is possible in the future we may want such other group versions, but this won't be in the scope of the current work.

Please be aware the groupversion field added here is NOT to indicate the group version of the provider API. Rather, it is used to indicate the group version of the DBaaS API the provider supports.

api/v1beta1/dbaasprovider_webhook.go
jianrongzhang89: Good idea. I added the default as part of the conversion webhook logic.

tchughesiv: hmm... should a default be added to v1beta1 schema as well? instead of conversion?
https://book.kubebuilder.io/reference/markers/crd-validation.html
```go
//+kubebuilder:default:=<any>
```

jianrongzhang89: Yes. Added the default.

api/v1beta1/dbaasprovider.go
tchughesiv: if we're setting a default and it appears to be a required field for functionality... should we make this required?

api/v1beta1/dbaasprovider_webhook.go
tchughesiv: if we're setting default and setting in conversion, is this block of code necessary?

api/v1beta1/dbaasprovider.go
tchughesiv: also, once required and with default being set... is the new conversion logic around this field necessary anymore?

tchughesiv (changes requested):

api/v1beta1/dbaasprovider.go
jianrongzhang89: Yes I tested it and found it is still required. This is because the conversion webhook essentially overrides the CRD validation and it it will end up the field to get set to empty if the conversion logic does not set the groupVersion field directly.

api/v1alpha1/dbaasprovider_conversion_test.go
```diff
@@ -43,25 +43,53 @@ var _ = Context("DBaaSProvider Conversion", func() {
...
 				Status: DBaaSProviderStatus{},
 			}
 			intermediate := v1beta1.DBaaSProvider{}
 			dst := DBaaSProvider{}
 
-			//intermediate.Hub()
 			Expect(src.ConvertTo(&intermediate)).To(Succeed())
 			Expect(dst.ConvertFrom(&intermediate)).To(Succeed())
-			Expect(dst).To(Equal(src))
+			assertProvidersEqual(&src, &dst)
```
tchughesiv, line 73: why this change?

api/v1alpha1/dbaasprovider_conversion.go
```diff
@@ -32,26 +48,81 @@ func (src *DBaaSProvider) ConvertTo(dstRaw conversion.Hub) error {
 
 	// Spec
 	dst.Spec.AllowsFreeTrial = src.Spec.AllowsFreeTrial
+	dst.Spec.GroupVersion = GroupVersion.String() // Set to v1alpha1 version
```
tchughesiv, line 51: we need a unit test check for this... at the very least verifying this is properly set in the "intermediate" v1beta1 object. ideally, we could see a reversed conversion where v1alpha1 becomes the "intermediate" and we don't lose any of the new fields on the other side. Maybe in the "src" object, this field is instead set to dbaas.group.com/v1beta1 ?

api/v1beta1/dbaasprovider.go
tchughesiv: thanks

tchughesiv (changes requested):
looking good! just a couple more things

config/default/manager-env-images.yaml
```diff
@@ -27,9 +27,9 @@ spec:
             - name: CSV_VERSION_MONGODB_ATLAS
               value: mongodb-atlas-kubernetes.v0.3.0
             - name: RELATED_IMAGE_COCKROACHDB_CATALOG
-              value: gcr.io/cockroach-shared/ccapi-k8s-operator-catalog:v0.0.4
+              value:  quay.io/jianrzha/ccapi-k8s-operator-catalog:v0.0.5-dev
```
xieshenzh, line 30: Are we supposed to use your image for now?

api/v1alpha1/dbaasprovider_conversion_test.go
jianrongzhang89: This is because in v1alpha1, Spec.InstanceParamterSpecs is an array. When the provider CR is converted to v1beta1 and then converted back, the sequence of the parameters in the array is not guaranteed to be the same, and consequently, using deepEqual to compare the 2 objects will fail.  I used assertProvidersEqual() to get around this issue by converting the array to a map and deep equal for go maps is not impacted the order of the elements.

tchughesiv: thanks

config/default/manager-env-images.yaml
jeremyary: good catch!

api/v1alpha1/dbaasprovider_conversion.go
jianrongzhang89: Good idea. Checking the intermediate object is helpful. We can not set it in the src object though, as the field does not exist in v1alpha1.

config/default/manager-env-images.yaml
jianrongzhang89: I reverted all references of devt images to existing release images. Once we have new provider images ready we will update the env config file to point to the new images.

tchughesiv: do we need to wait to merge this code? shouldn't an updated provider image be part of this PR?

tchughesiv (dismissed):
aside from provider image decision, lgtm

config/default/manager-env-images.yaml
jeremyary: my fault, didn't really think on implications here. Likely not a bad idea to keep your functional image until we get one from CRDB and then we can update with a quick PR after. Apologies!

jianrongzhang89: OK as we discussed lets use the devt build for CRDB image for now.

redhatHameed (approved):
lgtm

redhatHameed:
verified on cluster able create the DEDICATED successfully. 
```apiVersion: dbaas.redhat.com/v1beta1
kind: CrdbDBaaSInstance
metadata:
  resourceVersion: '11755382'
  name: ahameed
  uid: 1d107b84-e974-45c4-97f8-88c6c32ba062
  creationTimestamp: '2023-01-26T17:44:15Z'
  generation: 1
  managedFields:
    - apiVersion: dbaas.redhat.com/v1beta1
      fieldsType: FieldsV1
      fieldsV1:
        'f:metadata':
          'f:finalizers':
            .: {}
            'v:"crdbdbaasinstance.dbaas.redhat.com/cluster"': {}
          'f:ownerReferences':
            .: {}
            'k:{"uid":"9af56726-2f3c-4792-98e6-0c3065ae0e37"}': {}
        'f:spec':
          .: {}
          'f:inventoryRef':
            .: {}
            'f:name': {}
            'f:namespace': {}
          'f:provisioningParameters':
            .: {}
            'f:cloudProvider': {}
            'f:machineType': {}
            'f:name': {}
            'f:nodes': {}
            'f:plan': {}
            'f:regions': {}
            'f:storageGib': {}
      manager: manager
      operation: Update
      time: '2023-01-26T17:44:15Z'
    - apiVersion: dbaas.redhat.com/v1beta1
      fieldsType: FieldsV1
      fieldsV1:
        'f:status':
          .: {}
          'f:conditions': {}
          'f:instanceID': {}
          'f:instanceInfo':
            'f:createAt': {}
            'f:operationStatus': {}
            'f:regions.1.name': {}
            .: {}
            'f:updateAt': {}
            'f:cloudProvider': {}
            'f:creatorId': {}
            'f:numOfRegions': {}
            'f:cockroachVersion': {}
            'f:regions.1.sqlDns': {}
            'f:state': {}
            'f:plan': {}
          'f:phase': {}
      manager: manager
      operation: Update
      subresource: status
      time: '2023-01-26T18:46:31Z'
  namespace: openshift-dbaas-operator
  ownerReferences:
    - apiVersion: dbaas.redhat.com/v1beta1
      blockOwnerDeletion: true
      controller: true
      kind: DBaaSInstance
      name: ahameed
      uid: 9af56726-2f3c-4792-98e6-0c3065ae0e37
  finalizers:
    - crdbdbaasinstance.dbaas.redhat.com/cluster
spec:
  inventoryRef:
    name: my-db-provider-account-crdb
    namespace: openshift-dbaas-operator
  provisioningParameters:
    cloudProvider: AWS
    machineType: m5.large
    name: ahameed
    nodes: '3'
    plan: DEDICATED
    regions: us-east-1
    storageGib: '15'
status:
  conditions:
    - lastTransitionTime: '2023-01-26T18:22:46Z'
      message: Ready
      reason: Ready
      status: 'True'
      type: ProvisionReady
  instanceID: 50820275-03a8-4be7-af2c-b2ab37a68601
  instanceInfo:
    numOfRegions: '1'
    cockroachVersion: v22.2.3
    cloudProvider: AWS
    regions.1.sqlDns: ahameed-hcc.aws-us-east-1.cockroachlabs.cloud
    creatorId: ea1d2f94-9c86-488d-8c7d-563b77060632
    plan: DEDICATED
    createAt: '2023-01-26 17:44:15.690894 +0000 UTC'
    state: CREATED
    regions.1.name: us-east-1
    operationStatus: CLUSTER_STATUS_UNSPECIFIED
    updateAt: '2023-01-26 18:46:21.920796 +0000 UTC'
  phase: Ready```

PR #330, opened 2023-01-31
DBAAS-1170: workaround to the rds-controller upgrade issue for release 0.0.4

xieshenzh:
## Description

The aws `rds-controller` requires a configuration configmap and a credentials secret to finish install and upgrade.
The `rds-dbaas-operator` will scale it down to avoid this issue when installation, but during upgrade the `rds-dbaas-operator` is deleted. `rds-controller` will auto-upgrade,`rds-dbaas-operator` won't be installed until the `rds-controller` upgrade finishes or deleted. As a result, the upgrade will be in a dead lock if there is no rds provider account before upgrading.

The workaround is to create dummy aws secret and configmap if not exist during the `dbaas-operator` controllers installation. `ack-controller` and `rds-dbaas-operator` start upgrading after the `dbaas-operator` installation finishes.

## Verification Steps

Tested 3 scenarios:
1. fresh installation, create provider account for RDS, then create provider account for RDS
2. install v0.0.3, create provider account for RDS on v0.0.3, then upgrade to v0.0.4
3. install v0.0.3, not create provider account for RDS, then upgrade to v0.0.4 and create provider account for RDS

## Type of change

- [ ] Code Review completed

### DIFF
--- a/main.go
+++ b/main.go
@@ -124,6 +124,7 @@ func main() {
 		ClientDisableCacheFor: []client.Object{
 			&operatorframework.ClusterServiceVersion{},
 			&corev1.Secret{},
+			&corev1.ConfigMap{},
 		},
 	})
 	if err != nil {
--- a/api/v1beta1/webhook_suite_test.go
+++ b/api/v1beta1/webhook_suite_test.go
@@ -116,6 +116,7 @@ var _ = BeforeSuite(func() {
 		ClientDisableCacheFor: []client.Object{
 			&operatorframework.ClusterServiceVersion{},
 			&corev1.Secret{},
+			&corev1.ConfigMap{},
 		},
 	})
 	Expect(err).NotTo(HaveOccurred())
--- a/bundle/manifests/dbaas-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/dbaas-operator.clusterserviceversion.yaml
@@ -423,8 +423,6 @@ spec:
           verbs:
           - create
           - get
-          - list
-          - watch
         - apiGroups:
           - ""
           resources:
--- a/bundle/manifests/dbaas-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/dbaas-operator.clusterserviceversion.yaml
@@ -418,6 +418,16 @@ spec:
         - apiGroups:
           - ""
           resources:
+          - configmaps
+          - secrets
+          verbs:
+          - create
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - ""
+          resources:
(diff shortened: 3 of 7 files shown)


xieshenzh:
Overall coverage for the file is ~50%, but for the new code I added it is 73.3% . 

image

tchughesiv (commented):
thanks for the contrib! I have concerns about memory caching/leaks related to configmaps & secrets. Those are the 2 biggest offenders and we've had to take precaution to avoid CM watches in the past. With regard to secrets, while we do watch them... we take precautions like the following -
https://github.com/RHEcosystemAppEng/dbaas-operator/blob/346d6c647ba154d9c44effdfd4c0fd7abb726a5e/main.go#L124-L127

xieshenzh:
> thanks for the contrib! I have concerns about memory caching/leaks related to configmaps & secrets. Those are the 2 biggest offenders and we've had to take precaution to avoid CM watches in the past. With regard to secrets, while we do watch them... we take precautions like the following -
> 
> https://github.com/RHEcosystemAppEng/dbaas-operator/blob/346d6c647ba154d9c44effdfd4c0fd7abb726a5e/main.go#L124-L127
> 
> I worry, with a change like this... we'd have to worry about memory leaks again or, at the very least... this would prompt some intensive performance testing/sessions.

Are you suggesting we remove the watch permission and disable cache for configmap as well?

tchughesiv:
@xieshenzh at the very least, yes. we'll also need broader team discussion

tchughesiv:
@xieshenzh probably don't need the `list` perm either

main.go
```diff
@@ -124,6 +124,7 @@ func main() {
 		ClientDisableCacheFor: []client.Object{
 			&operatorframework.ClusterServiceVersion{},
 			&corev1.Secret{},
+			&corev1.ConfigMap{},
```
tchughesiv, line 127: please add this to the test suites as well...
here - 
https://github.com/RHEcosystemAppEng/dbaas-operator/blob/346d6c647ba154d9c44effdfd4c0fd7abb726a5e/controllers/suite_test.go#L124-L127
and here -
https://github.com/RHEcosystemAppEng/dbaas-operator/blob/346d6c647ba154d9c44effdfd4c0fd7abb726a5e/api/v1beta1/webhook_suite_test.go#L116-L119

tchughesiv (changes requested):

main.go
xieshenzh: Done

tchughesiv (approved):
lgtm

PR #338, opened 2023-02-07
DBAAS-1193: workaround to the ack-rds-controller v0.1.3 upgrade issue

xieshenzh:
## Description

With ack-rds-controller v0.1.3, the label selector of the deployment of the operator is changed. 
However, this field is immutable. 
As a result, upgrading ack-rds-controller from an older version to v0.1.3 fails.

Jira: https://issues.redhat.com/browse/DBAAS-1193

The workaround is to use the RHODA operator to check the status of the ack-rds-controller v0.1.3 csv. 
If the upgrade issue is detected, RHODA operator will delete the deployment of the ack-rds-controller.
Then the deployment of the ack-rds-controller will be re-created automatically, and the upgrade will contiune.

## Verification Steps

Upgrading RHODA v0.0.3 to 0.0.4.
Check if the upgrade succeeds and if the ack-rds-controller is upgraded to v0.1.3.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA or in issue number have been completed
- [ ] Unit tests added that prove the fix is effective or the feature works 
- [ ] Documentation added for the feature
- [ ] all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer

PR #343, opened 2023-02-10
DBAAS-1193: workaround to the ack-rds-controller upgrade issue

xieshenzh:
## Description

New upgrade issue was found with ARO cluster when the ack-rds-controller upgrades from v0.1.0 to v0.1.1.
Jira: https://issues.redhat.com/browse/DBAAS-1193

The new workaround is to delete the ack-rds-controller Subscription and CSV when dbaas-operator v0.0.4 is installed.
So that, ack-rds-controller v1.0.3 installation will be triggered when installing the rds-dbaas-operator, and ack-rds-controller upgrade is skipped.

## Verification Steps

Upgrading RHODA v0.0.3 to 0.0.4.
Check if the upgrade succeeds and if the ack-rds-controller is upgraded to v0.1.3.

## Type of change

- [ ] Code Review completed


bundle/metadata/dependencies.yaml
```diff
@@ -3,7 +3,3 @@ dependencies:
     value:
       packageName: rh-service-binding-operator
       version: ">=1.0.0"
-  - type: olm.package
-    value:
-      packageName: ack-rds-controller
-      version: ">=0.0.27 <=0.1.3"
```
redhatHameed, line 9: is this intentional removed ? if yes how the RDS controller will be installed ?

xieshenzh: It will be installed as a dependency of the rds-dbaa-operator.

tchughesiv: this isn't good enough... in my prior testing its required in both places. If nothing else, dbaas-operator at the very least is where this dep needs to be. This is because during a RHODA upgrade, if ack-rds-controller has since release a new version... say `0.2.0`, ACK will upgrade without this dep declared.

tchughesiv: this should not be removed

tchughesiv (changes requested):

controllers/dbaasplatform_controller.go
```diff
@@ -88,7 +86,7 @@ type DBaaSPlatformReconciler struct {
 //+kubebuilder:rbac:groups=dbaas.redhat.com,resources=*/finalizers,verbs=update
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=catalogsources;operatorgroups,verbs=get;list;create;update;watch
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=subscriptions,verbs=get;list;create;update;watch;delete
-//+kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=get;update;delete
+//+kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=get;update;delete;list
```
tchughesiv, line 89: this could be dangerous... similar to the configmap/secret memory leak issues... CSVs are large in size and number in big clusters like sandbox. If we can avoid adding list privilege here... we need to. I'm not sure we can add CSV to `ClientDisableCacheFor`, it may affect existing rhoda code.

tchughesiv (changes requested):

bundle/metadata/dependencies.yaml
xieshenzh: The auto-upgrade of the ack-rds-controller is [broken](https://github.com/RHEcosystemAppEng/dbaas-operator/pull/338). Since the upgrade of the ack-rds-controller is not maintained, it is not necessary to allow it to upgrade, which causes various issues.
Changes in this PR will delete any ack-rds-controller as soon as the dbaas-operator is installed in the namespace, upgrade of the ack-rds-controller will not take place during RHODA upgrade.
When the rds-dbaas-operator is re-installed, the properly version of the ack-rds-controller will be installed as a dependency.

tchughesiv: this is the benefit of using olm deps

xieshenzh: If the ack-rds-controller is installed in another namespace, rhoda will not work. Because the secret and configmap of the ack-rds-controller must be set by the rhoda operator in the rhoda installation namespace in order to create a provider account, then rhoda will scale up the deployment of the ack-rds-controller in the rhoda installation namespace.

tchughesiv: but, wouldn't anyone who's installed ack in another namespace have configured it already? why would we care if it's in same namespace?

xieshenzh: The configuration information of the ack-rds-controller should be provided via rhoda provider account creation. Only one provider account can be created for RDS in a cluster, and the credentials of the provider account is used to set up the ack-rds-controller.
If users configure and install the operator in another namespace without using rhoda, then it is supposed to work independent of rhoda.

controllers/dbaasplatform_controller.go
```diff
@@ -433,82 +425,40 @@ func (r *DBaaSPlatformReconciler) fixConversionWebhooks(ctx context.Context) err
...
+		for _, crd := range csv.Spec.CustomResourceDefinitions.Owned {
+			if crd.Name == "dbinstances.rds.services.k8s.aws" && crd.Kind == "DBInstance" {
+				instanceCRD = true
+			} else if crd.Name == "dbclusters.rds.services.k8s.aws" && crd.Kind == "DBCluster" {
+				clusterCRD = true
 			}
-			if err := cli.Create(ctx, cm); err != nil {
+		}
+		if instanceCRD && clusterCRD {
+			if err := cli.Delete(ctx, &csv); err != nil {
```
tchughesiv, line 457: What happens when the dbaas-manger pod gets restarted? Would this cause the ack sub to be deleted?

xieshenzh: Thanks. The ack-rds-controller will be deleted in this case.
Changed the code to also delete rds-dbaas-operator when starting dbaas-operator. Then the rds-dbaas-operator and ack-rds-controller will be re-installed when re-starting dbaas-operator.

tchughesiv: I'm not sure that's a solution though. We can't have ack and rds being uninstalled/reinstalled every time the pod restarts... which could be often.

xieshenzh: I could check the version of the ack-rds-controller only delete it when the version is not right.
Otherwise, I am not sure if there is a good solution to the ack-rds-controller upgrade issues.

PR #354, opened 2023-03-13
[DBAAS-1255] community bundle changes

tchughesiv:
start w/ new catalog image for first community release.
removing mongo atlas from platform installer since an official version already exists in the community operator hub.
modifying SBO dependency to support either the RH version or community version


jianrongzhang89:
@tchughesiv we also need to build a new release of RDS operator as it has been rebranded and update the env config file.

tchughesiv:
> @tchughesiv we also need to build a new release of RDS operator as it has been rebranded and update the env config file.

That can be done in a separate pull request

